PR #13, opened 2024-04-05
Switch Tuning database to SQLite3 for Incremental Tuning

xinyazhang:
## Motivation

Old approach only writes a single giant `.json​` file as the tunig database at the end of completing all tuning tasks, which surely won't scale.

New approach introduces a SQLite management script call `table_tool.py​` to update a monolithic SQLite(3) database file `v2python/rules/tuning_database.sqlite3`

Fundamentally this database is a fancier version of CSV files, but we want a few features from database like integrity (i.e., no broken CSV lines), etc.
MongoDB fits this purpose better, but it has licensing issues (Server Side Public License is prohibited) and introduces yet another dependency, while SQLite is public domain, works well enough, and is part of Python Standard library.

## Changes
1. A `table_tool.py​` script that manages SQLite database file according to schemaless JSON inputs.
2. Updated `tritonsrc/tune_flash.py` to update tuning database in a more progressive way. The new tuning script also drops the dependency of pytest framework.
3. Refactored `KernelTuningDatabaseForArch` class hierarchy
    * Reading of SQLite database implemented by `SQLiteKernelTuningDatabaseForArch` in  `v2python/sqlite_tuning_database.py`
4. Replace the json tuning database with `v2python/rules/tuning_database.sqlite3`
5. 'STAGE: tl.constexpr' -> 'CAUSAL: tl.constexpr' in `attn_fwd` Triton kernel. This simplifies the tuning database schemas since all Flash Attention kernels now use the same columns to describe the inputs


v2python/tuning_database.py
```diff
@@ -1,284 +1,69 @@
...
-                # print(f'{eti=}')
-                if eti['tuned_kernel'] == ti['tuned_kernel'] and eti['compiler_options'] == ti['compiler_options']:
-                    is_dup = True
-                    break
-            if not is_dup:
-                self._index_dedup[tup].append(ti)
-        if False:  # debug
-            tup=('*fp16:16', 1, 16, True, True)
-            print(f'_build_db_index {self._index[tup]=} {self._index_dedup[tup]=}')
+        assert False
```
groenenboomj, line 22: Isn't this a dead method now?  Can it be removed? Same comment for other functions that are just 'assert False' now.

tritonsrc/attn_torch_function.py
```diff
@@ -51,49 +52,43 @@ def tuned_attn_fwd(
...
-    max_seqlens_q, max_seqlens_k,
-    head_dim_q, head_dim_k,
+    seqlen_q,
+    seqlen_k,
+    head_dim,
     dropout_p,
     philox_seed,
     philox_offset_base,
     encoded_softmax,
-    VARLEN: tl.constexpr,
```
groenenboomj, line 62: I take it this will be brought back or addressed on the next kernel merge?

v2python/downgrader.py
```diff
@@ -0,0 +1,49 @@
+# Copyright © 2023-2024 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+'''
+Used in conjunction with PARTIALLY_TUNED_FUNCTIONALS
+
+Commonly enabling functionals will cost extra resources,
+and thus make the fallback turing information unusable
```
groenenboomj, line 8: Could you elaborate on what this enables?  Was this the answer to tuning regressions?

groenenboomj (commented):
How has the manual effort in  refining the solution changed with these updates? Any impact?

v2python/tuning_database.py
xinyazhang: This is an abstract method in the base class and must have some definition.

tritonsrc/attn_torch_function.py
xinyazhang: Yes. For now it's no point to keep parameters that does not exist in the Triton kernel.

v2python/downgrader.py
xinyazhang: This allows `KernelDescription` objects to "downgrade"/"patch" some tuning parameters according to enabled functionals.

This function is not used in current builds, and was introduced because Triton/tritonsrc combinations, the maximal `(BLOCK_M,BLOCK_N)` is `32,16` when dropout becomes true, and at that time the tuning parameters has to be specified by `KernelDescription` class.

Now the selection process is largely handled by the tuning database, but the `DOWNGRADER` member is still supported by `KernelDescription` class for possible future use.

groenenboomj (approved):
All feedback addressed.

PR #14, opened 2024-04-16
Add matrix bias to forward/backward kernel

xinyazhang:
All UTs Passed. No performance regression on existing forward kernel cases (i.e., `test/performance_forward.py`).


tritonsrc/bwd_split_kernel.py
```diff
@@ -102,6 +109,34 @@ def bwd_kernel_dk_dv(
...
+                block_shape=(BLOCK_M, BLOCK_N),
+                order=(1, 0)
+                )
+        if (stride_dbz == 0 or stride_dbh == 0) or stride_dbm == 0:
+            store_db = False
+        else:
+            store_db = True
+        # Still have to make one even if no_db = False
+        # due to a limit of Triton: runtime branches must have identical data types.
+        DB_block_ptr = tl.make_block_ptr(
```
groenenboomj, line 130: I see a store but no load so in the end what is this storing?

groenenboomj (commented):
I'm trusting the database changes but I was also concerned to not see testing above 2k sequence lengths as most use cases where we are beneficial are higher than that.

tritonsrc/bwd_split_kernel.py
```diff
@@ -170,6 +212,24 @@ def bwd_kernel_dk_dv(
...
+            if q_padded and k_padded:  # CAVEAT: using "or" disables the partial boundary_check branches
+                bias = tl.load(B_block_ptr, boundary_check=(0,1), padding_option="zero")
+            elif q_padded:
+                bias = tl.load(B_block_ptr, boundary_check=(0,), padding_option="zero")
+            elif k_padded:
+                bias = tl.load(B_block_ptr, boundary_check=(1,), padding_option="zero")
+            else:
+                bias = tl.load(B_block_ptr)
+            """
+            bias = tl.load(B_block_ptr, boundary_check=(0,1), padding_option="zero")
```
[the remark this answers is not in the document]
xinyazhang: I haven't check if it's a compiler problem, and even if it is we do not have the capacity to fix it.
The performance is not a concern before we can make it work.

tritonsrc/bwd_split_kernel.py
xinyazhang: D is the prefix for differentials, and DB is the differential of bias tensor.

xinyazhang:
> I'm trusting the database changes but I was also concerned to not see testing above 2k sequence lengths as most use cases where we are beneficial are higher than that.

Any sequence longer than 2k will be treated as 2k when looking up the optimal kernel. I don't think the optimal compiler options will change for longer sequences.

tritonsrc/bwd_split_kernel.py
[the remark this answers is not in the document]
groenenboomj: Sounds good then.  Just make an issue so we don't lose track. We'll need to root cause this before we hit more boundary issues.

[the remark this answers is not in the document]
xinyazhang: Filed as https://github.com/ROCm/aotriton/issues/17

tritonsrc/attn_torch_function.py
```diff
@@ -474,6 +466,8 @@ def backward(ctx, do, _, fwd_tuning_result):
         else:
             BLOCK_M = 128
             BLOCK_N = 64
+        if q.dtype == torch.float32:
+            BLOCK_M //= 2
```
groenenboomj, line 470: I take it this was from memory limitations?

xinyazhang: Yes, but unfortunately this doesn't fix all problems.
fp32 gets broken in other places.

groenenboomj (approved):
Chatted offline, looks good.

PR #19, opened 2024-04-23
Add new triton kernel debug_fill_dropout_rng

xinyazhang:
This is needed for `at::native::_fill_mem_eff_dropout_mask_`

All newly added UTs in `test_forward.py` passed. No regressions.


groenenboomj (approved):
With the explanation in person this seems reasonable.  We should talk about delineating the memory efficient and flash attention specific APIs in a later round of changes.

PR #22, opened 2024-04-30
Add FP32 and Bias to fulfill the functionalities required by `torch.nn.attention.SDPBackend.EFFICIENT_ATTENTION`

xinyazhang:
This PR includes the following major changes

1. Add Bias support in the Triton kernel, for both forward and backward directions
2. Add `fp32` datatype support, and the corresponding tuning database information
3. Fix "argument list too long" error during linking
4. Improved `table_tool.py` to partially dump/load `.csv` file, allowing database merging
5. Refactor the UT to use PyTorch's method to estimate ATOL/RTOL

Known limitations:
1. Bias assumes real Rank 4 tensor (`.expand()`-ed ones are unlikely to work) for backward direction. No checking is performed on this requisite and failure may be silent
2. `test_forward.py` is still using the old method to estimate ATOL/RTOL


tritonsrc/attn_torch_function.py
```diff
@@ -471,8 +471,8 @@ def backward(ctx, do, _, fwd_tuning_result):
             BLOCK_M = 128
             BLOCK_N = 64
         if q.dtype == torch.float32:
-            BLOCK_M //= 2
-            BLOCK_N //= 2
+            BLOCK_M = max(16, BLOCK_M // 2)
```
groenenboomj, line 474: Is this just for FP32?  Could we gate on that dtype?

v2python/rules/tuning_database.sqlite3
groenenboomj, line 1: Was this just from adding the new features or adding new architectures?

groenenboomj (commented):
What are the perf impacts of these changes?

v2python/rules/tuning_database.sqlite3
xinyazhang: Yes. (actually I'm considering switch to .csv format for readability...)

tritonsrc/attn_torch_function.py
xinyazhang: Just for fp32. Otherwise fp32 result will fail.

groenenboomj (approved):

PR #23, opened 2024-05-08
Resolve cmake conflicts when adding aotriton into TE via add_subdirectory

wangye805:
change CMAKE_SOURCE_DIR to CMAKE_CURRENT_SOURCE_PARENT_DIR since other repo may add aotriton as a dependency via add_subdirectory


xinyazhang (approved):
LGTM.

PR #24, opened 2024-05-14
[mGPU] Run hipModuleLoadDataEx for each GPU device.

xinyazhang:
Fix `TritonKernel::invoke` for multi-GPU scenarios.

`hipModuleLoadDataEx` only loads the image into a module that is specific to current GPU. Hence if another `TritonKernel::invoke` was made on a different GPU, the function call will fail since the module is only valid for the previous GPU.

### DIFF
--- a/include/aotriton/_internal/triton_kernel.h
+++ b/include/aotriton/_internal/triton_kernel.h
@@ -11,6 +11,9 @@
 
 #include "../runtime.h"
 #include <vector>
+#include <unordered_map>
+#include <shared_mutex>
+#include <tuple>
 
 namespace aotriton {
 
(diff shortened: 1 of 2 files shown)


include/aotriton/_internal/triton_kernel.h
```diff
@@ -24,11 +27,24 @@ class TritonKernel {
...
+  hipFunction_t cfind_function(int device_id) const;
+
   const void* kernel_image_ = nullptr;
   size_t image_size_ = 0;
+  struct DeviceFunction {
+    DeviceFunction(int device_id_,
+                   hipModule_t mod_,
+                   hipFunction_t func_);
+    ~DeviceFunction();
+    int device_id = -1;
```
wangye805, line 40: This device_id is not used later on. And you already have a unordered_map whose key is device_id

include/aotriton/_internal/triton_kernel.h
```diff
@@ -24,11 +27,24 @@ class TritonKernel {
...
+    hipModule_t mod = nullptr;
+    hipFunction_t func = nullptr; 
+  };
+  std::unordered_map<int, DeviceFunction> funcache_;
```
wangye805, line 44: funcache_ --> fun_cache_, just to be consistent with your own coding style

wangye805 (approved):
LGTM with minor comments

include/aotriton/_internal/triton_kernel.h
xinyazhang: It's kept here in case of possible sanity checks.

include/aotriton/_internal/triton_kernel.h
xinyazhang: Oh this is a similar to trick of "AOT Triton -> AOTriton", and "Func cache -> Funcache"

wenchenvincent (approved):
LGTM.

PR #29, opened 2024-05-31
Refactor the build system

xinyazhang:
Use `CMakeLists.txt` instead of `Makefile` generated by `v2python/generate_shim.py` and `v2python/generate_compile.py` to process the build. Also do not overwrite generate source files if no changes were found.

Now AOTriton can avoid most duplicated work during the re-build.

This PR also fixes the false negative UT failures in main introduced in 0.5b.


xinyazhang:
Have to merge by force since Joseph is OOO till the end of this week.

PR #33, opened 2024-06-19
Install .so if AOTRITON_NO_SHARED is OFF

jithunnair-amd:
Needed to move PyTorch builds to use AOTriton as shared library


xinyazhang (approved):
LGTM

PR #42, opened 2024-09-09
Add cmake option AOTRITON_NAME_SUFFIX to resolve name conflicts

xinyazhang:
This adds suffix to namespace `aotriton` and the library file prefix `libaotriton`. For example, build with `-DAOTRITON_NAME_SUFFIX=123` will generate the library with `namespace aotrition123` and `libaotriton123_v2.so`.

This is to address potential name conflicts with PyTorch since PyTorch will bundle AOTriton shared library starting 2.4+

### DIFF
--- a/bindings/module.cc
+++ b/bindings/module.cc
@@ -1,7 +1,7 @@
 // Copyright © 2023-2024 Advanced Micro Devices, Inc.
 // SPDX-License-Identifier: MIT
 
-#include <aotriton/aotriton_config.h>
+#include <aotriton/config.h>
 #include <aotriton/dtypes.h>
 #include <aotriton/flash.h>
 #include <aotriton/runtime.h>
--- a/bindings/module.cc
+++ b/bindings/module.cc
@@ -1,6 +1,7 @@
 // Copyright © 2023-2024 Advanced Micro Devices, Inc.
 // SPDX-License-Identifier: MIT
 
+#include <aotriton/aotriton_config.h>
 #include <aotriton/dtypes.h>
 #include <aotriton/flash.h>
 #include <aotriton/runtime.h>
@@ -11,6 +12,9 @@
 #include <string>
 
 namespace py = pybind11;
+#if AOTRITON_ENABLE_SUFFIX
+namespace aotriton = AOTRITON_NS;
+#endif
 
 namespace pyaotriton {
   namespace v2 {
--- a/bindings/module.cc
+++ b/bindings/module.cc
@@ -221,6 +221,17 @@ namespace pyaotriton {
       .def_property_readonly("strides", &aotriton::TensorView<0>::strides)
       .def_property_readonly("data_ptr", &aotriton::TensorView<0>::data_ptr)
       .def_property_readonly("dtype", &aotriton::TensorView<0>::dtype);
+    m.def("get_name_suffix",
+#if AOTRITON_ENABLE_SUFFIX
+#define xstr(s) str(s)
+#define str(s) #s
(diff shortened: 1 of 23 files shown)


bindings/module.cc
```diff
@@ -217,6 +221,17 @@ namespace pyaotriton {
       .def_property_readonly("strides", &aotriton::TensorView<0>::strides)
       .def_property_readonly("data_ptr", &aotriton::TensorView<0>::data_ptr)
       .def_property_readonly("dtype", &aotriton::TensorView<0>::dtype);
+    m.def("get_name_suffix",
+#if AOTRITON_ENABLE_SUFFIX
+#define xstr(s) str(s)
+#define str(s) #s
+          []() -> std::string { return xstr(AOTRITON_NAME_SUFFIX); }
```
jithunnair-amd, line 228: @xinyazhang I'm sure there's some nuance with the preprocessing flow here, but can you elaborate why we need two macros `xstr` and `str`?

v2src/CMakeLists.txt
```diff
@@ -206,8 +215,10 @@ install(FILES "${AOTRITON_V2_BUILD_DIR}/${AOTRITON_LIBRARY_FILE}" DESTINATION ${
...
-add_dependencies(aotriton aotriton_v2)
+# add_dependencies(aotriton aotriton_v2)
 # target_link_libraries(aotriton INTERFACE ${CMAKE_INSTALL_PREFIX}/lib/libaotriton_v2.a)
 # target_include_directories(aotriton INTERFACE ${CMAKE_INSTALL_PREFIX}/include)
-target_link_libraries(aotriton INTERFACE ${AOTRITON_V2_BUILD_DIR}/${AOTRITON_LIBRARY_FILE})
-target_include_directories(aotriton INTERFACE ${CMAKE_CURRENT_SOURCE_PARENT_DIR}/include)
+# target_link_libraries(aotriton INTERFACE ${AOTRITON_V2_BUILD_DIR}/${AOTRITON_LIBRARY_FILE})
+# target_include_directories(aotriton INTERFACE ${CMAKE_CURRENT_SOURCE_PARENT_DIR}/include)
+# target_include_directories(aotriton INTERFACE ${CMAKE_CURRENT_SOURCE_PARENT_DIR}/include)
+target_link_libraries(aotriton INTERFACE aotriton_v2)
```
jithunnair-amd, line 224: @xinyazhang Are these changes related to the name suffix feature? Can you please explain this changeset?

bindings/module.cc
xinyazhang: @jithunnair-amd https://gcc.gnu.org/onlinedocs/cpp/Stringizing.html

v2src/CMakeLists.txt
xinyazhang: the suffix is not long added to the symbols, but also added to the output files. Linking `aotriton INTERFACE` can eliminate the necessary to construct the output file name with suffix again.

jithunnair-amd:
Build was successful and produced following outputs:
```
#ls -lart /usr//lib/aotriton/lib/
total 274288
-rw-r--r-- 1 root root 280861184 Sep 27 20:40 libaotritonmy_suffix_v2.so

#grep suffix /usr//lib/aotriton/include/aotriton/ -r
/usr//lib/aotriton/include/aotriton/config.h:#define AOTRITON_NAME_SUFFIX my_suffix
/usr//lib/aotriton/include/aotriton/config.h:#define AOTRITON_NS aotritonmy_suffix
```

jithunnair-amd (approved):

PR #49, opened 2024-10-08
GQA Support

xinyazhang:
Add GQA Support to AOTriton

There is no API change. Number of heads have already been provided by 2nd dimension of the input Q/K/V tensors


vgokhale (approved):
LGTM!

PR #52, opened 2024-11-01
libaotriton_v2.so: Fix 'argument list too long" error

prarit:
This error is output when building the libaotriton_v2.so:

make: /bin/sh: Argument list too long
make: *** [Makefile.shim:7976: libaotriton_v2.so] Error 127
make: *** Waiting for unfinished jobs....

This can be resolved by using the ar.txt file that contains a list of the object files instead of adding them individually as arguments.

Signed-off-by: Prarit Bhargava <prarit@redhat.com>


xinyazhang:
@prarit Thanks for your contribution!
However, `Makefile` based build system has been deprecated since AOTriton 0.6b by https://github.com/ROCm/aotriton/pull/29

Current build system is purely CMake based and does not suffer from this problem.

PR #56, opened 2024-11-14
Port fixes from 0.7.2b

xinyazhang:
## Major Changes

1. varlen support fixes
2. Fix the numerical error due to rounding differences b/w FMA and MUL+SUB
    + The test is added `test_large_bf16_nan_values` in `tritonsrc/test_backward.py`

Note: "Fix NaN created by 0.0 (from sm_scale) * -inf (from masking)." is
firstly developed on main and then ported to 0.7.2b


xinyazhang:
Close due to 0.7.2b fix adds new bug. Will include these in the new PR

PR #84, opened 2025-03-24
Cut final binary size

xinyazhang:
## Summary

This PR reduces the `libaotriton_v2.so` from 49 MiB to 9.6 MiB (Release build, all default arches in 0.9b).

## Major Changes

* [shim] share common substring of TritonKernel's stem_name
* [shim] remove `#include <iostream>`
* [build] change default visibility to "hidden"
* [shim] deduplicate the binning lambda functions
* [shim] replace std::function with function pointers
* [shim] Store autotune strings compactly and re-write the TritonKernel initialization with loop
* [shim] re-order the perf fields with sizes.

## Minor Changes

* [v2python] move build_object_file_description to ObjectFileDescription
* [doc] update the build instruction to use `ninja install/strip`
* LICENSE: Update Copyright years

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ cd build
 export PKG_CONFIG_PATH="${PKG_CONFIG_PATH}:${CONDA_PREFIX}/lib/pkgconfig"
 cmake .. -DCMAKE_INSTALL_PREFIX=./install_dir -DCMAKE_BUILD_TYPE=Release -DAOTRITON_GPU_BUILD_TIMEOUT=0 -G Ninja
 # Use ccmake to tweak options
-ninja install
+ninja install/strip  # Use `ninja install` to keep symbols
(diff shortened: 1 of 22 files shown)


pruthvistony:
@xinyazhang ,
I read the comment explaining the breakup of kernel names into "KERNEL NAME STORAGE SCHEME", but I dont understand the algorithm. May be talk on how it is done will help.

pruthvistony:
> ~This is WIP since I haven't validate if the library works at runtime.~
> 
> Confirmed it works image (No need to validate with full suite, the main concern of this PR is the shim layer cannot dispatch the workload to GPU kernel)

I dont understand what do you mean here? Any you explain more.

jithunnair-amd (approved):
looks okay from what I can understand :)

PR #85, opened 2025-03-25
Add pytest-xdist support to test/test_backward.py

xinyazhang:
This allows us run unit tests on all GPUs in parallel

Check `docs/How To Run Tests.md` for how to run tests.


test/test_backward.py
```diff
@@ -16,6 +16,10 @@
 )
 from _common_test import SdpaContext, SdpaParams, SdpaContextFromNPZ, AOTRITON_TORCH_ONLY_USE_CPU
 
+@pytest.fixture()
+def torch_gpu(worker_id):
+    return int(worker_id[2:]) if worker_id != "master" else None
```
jithunnair-amd, line 21: What is a typical value of `worker_id`?

xinyazhang: Very weird values and not well documented. It is `gw0` `gw1` etc.

docs/How To Run Tests.md
```diff
@@ -0,0 +1,26 @@
...
+
+# Pre-requisites
+
+* `pip install -r requirements-dev.txt`
+
+# Pre-requisites for parallel testing on multi-GPUs
+
+* AOTriton >= 0.10.0
+* `jq` (usually available as `apt install jq` or `dnf install jq`)
+* Python CLI `amd-smi`. Install with https://rocm.docs.amd.com/projects/amdsmi/en/latest/install/install.html#manually-install-the-python-library if not available in current python venv.
```
jithunnair-amd, line 26: Do you use the *amdsmi python package* or just the `amd-smi` ROCm package? Based on `NGPUS=$(amd-smi list --json|jq length)`, it looks like it's only the latter?

jithunnair-amd (approved):
LGTM

docs/How To Run Tests.md
xinyazhang: `amd-smi-lib` provides the CLI tool `amd-smi` which symlinks to python script `/opt/rocm-X.Y.Z/libexec/amdsmi_cli/amdsmi_cli.py`. This script still requires `amdsmi` **python** package.

PR #93, opened 2025-04-24
template/autotune_table_entry.cc: include iostream for cerr

scottt:
Building with g++-14.2 from Fedora 42 was failing with:

```
(...)/aotriton/build/v2src/flash/autotune.bwd_preprocess/FONLY__^bf16@16,80,True___gfx1151.cc: In function ‘void aotriton::v2::flash::autotune::Autotune_bwd_preprocess__A0__F33(aotriton::v2::flash::BwdPreprocessParams&, int)’:
(...)/aotriton/build/v2src/flash/autotune.bwd_preprocess/FONLY__^bf16@16,80,True___gfx1151.cc:95:10: error: ‘cerr’ is not a member of ‘std’
   95 |     std::cerr << __FILE__ << " kernel_index = " << int(kernel_index) << std::endl;
      |          ^~~~
(...)/aotriton/build/v2src/flash/autotune.bwd_preprocess/FONLY__^bf16@16,80,True___gfx1151.cc:8:1: note: ‘std::cerr’ is defined in header ‘<iostream>’; this is probably fixable by adding ‘#include <iostream>’
    7 | #include <aotriton/cpp_tune.h>
  +++ |+#include <iostream>
    8 | #include <string_view>
```


v2src/template/autotune_table_entry.cc
```diff
@@ -5,6 +5,7 @@
 #include <aotriton/_internal/triton_kernel.h>
 #include <aotriton/_internal/kernel_cluster.h>
 #include <aotriton/cpp_tune.h>
+#include <iostream>
```
xinyazhang, line 8: Need to guard this with `#ifndef NDEBUG`-`#endif`

This was removed due to extra static initialization overhead by pulling into this header file.
The content of this template file will be duplicated into about 9000+ files (depending on the number of arch to support) and any small overheads will eventually be significant.

Otherwise LTGM

xinyazhang (changes requested):
Need to guard with #ifndef NDEBUG

xinyazhang (approved):

PR #95, opened 2025-05-13
Dispatcher V3 Part 3: Operators

xinyazhang:
# Overview

This PR rewrites the dispatcher and code generation, and now the new
dispatcher can be formally called as "V3".

## API Changes

* [v3] Add new namespace `aotriton::v3`, and exposed as `pyaotriton.v3`
* [v3] V3 API pattern follows `API(Param, ParamVersion, hipStream_t, Options)`
  - i.e., all arguments are passed with a C struct
  - ParamVersion is passed for possible backward compatibility support
  - CAVEAT: Empty tensors should be passed with `&TensorView<Rank>()` (which
    create a TensorView with `.data_ptr() == nullptr`. Passing
    `(TensorView*)nullptr` causes segfault in HIP Runtime.
* [v3] APIs added in `aotriton::v3` will be referred as `operators`
  - An operator can be implemented by multiple GPU kernels
  - An operator can have different implementations (referred as "backends")
  - An operator can be tuned to select optimal backend
* [v3] backends and GPU kernels of an operator share input/output parameters
  - It is possible to copy and paste from the `Operator` class
  - However, `SHARED_IFACE` is more preferred to avoid duplicated definition
* [v2] Remove deprecated API `debug_fill_dropout_rng` and
  `debug_fill_dropout_rng_tensor`

## Major Changes

* [gen] Move all python code to `v3python`
  - They are refactored into multiple sub-packages.
* [gen] Use `pandas` for database access
* [gen] Explicitly typed parameter system
  - All `*_CHOICES` values are now parsed into `v3python.base.TypedChoice`
    objects for code generation
  - Conditional Values now are extended from `TypedChoice`
  - In v2 they are python built-in values
  - See `v3python/base/README.md` for more details
* [shim] Move all C++ code to `v3src`
* [shim] Code template moved to `v3python/codegen/template`
* [storage] New hsaco name format
  - True and False in `F__` section is replaced with `T` and `F`
  - `^` was replace with full-width asterisk `＊`
  - UTF-8 is assumed and enforced wherever possible
  - This is considered as "Major change" because of the newly added UTF-8
    assumption.
* [gen] Add "Metro Kernel" which represent a sequence of kernels to implement
  an operator
* [gen] Add "Conditional Kernel" to Metro kernel
  - "Returned encoded softmax" is supported with conditional kernel:
    "if `encoded_softmax` then `debug_simulate_encoded_softmax`"
  - Varlen is supported through conditional kernel: "if varlen then
    `bwd_preprocess_varlen` else `bwd_preprocess`"
* [test] Add v3 API tests. Enabled with env var `V3_API=1`

## Minor Changes

* [build] `ccache` will be used when detected
* [gen] Unify `generate_shim` and `generate_compile` to `generate`
* [kernel] Rename arguments to ensure kernel arguments within the same operator share
  the name.
  - This is optional for Triton source, the actual name is encoded in `rules/`
    classes, not in Triton source.

## Known Problems

* [db] Operator's tuning database is still under consideration
  - This PR put them as a separate table tentatively
  - However it is more attractive to put them as part of autotune's table
* [gen] Generator for Operator's tuning (`optune` in the code) is incomplete


xinyazhang:
No regression in V3 API
image
(64 failures are existing failures in V2)

jeffdaily (approved):

PR #96, opened 2025-05-22
Re-implement Causal Masks with Windowed Attention

xinyazhang:
# Overview

This PR implements both causal variants with Sliding Window Attention
(referred as "Windowed attention" in the following)

For the two causal variants, see https://github.com/pytorch/pytorch/issues/108108

## API Changes

Windowed attention is exposed through V3 API.
V2 API is unmodified for backward compatibility.

* [v3]: Support Windowed Attention in `attn_fwd_params`/`attn_bwd_params`
  + Designate `causal_type = 3` as windowed attention
  + Add `window_left` and `window_right` fields
  + Add two special values `WindowValue::TopLeftAligned` and
    `WindowValue::BottomRightAligned` to support varlen.
* [v3] Remove `CausalType::TopLeftAligned` and
  `CausalType::BottomRightAligned` constants since they will not be compiled
  into the GPU kernels.
  
## Major Changes
  
* [kernel] Unify the causal argument as `CAUSAL_TYPE` and support 0/1/2/3 as
  none/topleft/bottomright/windowed.
* [kernel] re-design and unify the calculation of masked/full blocks.
  Implemented in `masked_load_store` as `parse_window` and
  `calculate_intervals`
* [kernel] All FA inner kernels now supports calculation over at most two
  discontinuous sections, to avoid calling the internal kernel twice for
  windowed attention.
  + Consequently `composed_load_with_offset` replaces `composed_advance` in
    the kernel since skipping blocks is needed if there are two sections.
  + Block pointers are replaced by conventional pointers and
    `composed_load_with_offset`
* [kernel] Swap batch and block grid index according to
  https://github.com/ROCm/aiter/pull/307

## Minor Changes

* [v3] Change `enum class` to `static constexpr` in `struct`.
  + Otherwise compiler complains: "cannot convert `WindowValue` to `int32_t`
    {aka `int` } in initialization"
* [test] Format `D_HEAD` by adding `hdim` prefix
  + This makes matching hdim only (like `-k 'CausalOn and hdim16-'`) possible.
* [tritonsrc] Support `SKIP_DK_DV` `SKIP_DQ` `SKIP_DB` env var in testing
* [test] re-purpose `causal` in `attention`, to support
  + `bool`: None/CausalType::TopLeftAligned
  + `CausalType.NONE/CausalType.TOP_LEFT/CausalType.BOTTOM_RIGHT`
  + tuple: `(window_left, window_right)`
* [LICENSE] Update Copyright years

## Known Problems

* [test] Due to lack of native support of Windowed Attention in PyTorch, this
  PR is not aiming to add fully covered windowed attention testing. For now
  the test suite only validate the windowed attention implementation can fully
  simulate top-left aligned causal mask.
  + It is possible to use `attn_mask` parameter supported by
    `torch.nn.functional.scaled_dot_product_attention` to simulate windowed
    attention. However, "how do we know `attn_mask` tensor fully matches
    sliding windowed attention syntax in other implementations" becomes an
    unsolved question, i.e., the generation `attn_mask` itself needs another
    set of tests.
* [tritonsrc] grid calculator in `attn_torch_function` is not updated with the
  kernel, will be fixed by another PR.


xinyazhang:
image
All (hdim < 256) UTs passed. For hdim=408/512, speculating register pressures and can be fixed with updated tuning database.

binding7012 (approved):

pruthvistony (approved):

PR #99, opened 2025-06-10
Fix: correct Python executable path for venv package discovery

### LINKED ISSUE #98: [Issue]: Python package discovery in venv during AOTriton build
When building AOTriton, the CMakeLists.txt files directly reference ${VENV_DIR}/bin/python for executing Python commands within the virtual environment. This can lead to issues with Python package discovery, especially if the python executable itself isn't directly in the bin directory or if system-wide Python dependencies are expected to be available through the standard PATH.

Credits go to my colleagues for finding this issue:
- Prarit Bhargava <prarit@redhat.com>
- Ali Raza <araza@redhat.com>

EmilienM:
This patch updates the CMakeLists.txt files in aotriton and v2src to correctly
locate the Python executable within the virtual environment.

Previously, the build process explicitly called `${VENV_DIR}/bin/python`.
This could lead to package discovery issues if the `python` executable
wasn't directly in that path, or if system-wide Python dependencies were implicitly relied upon.

The updated commands now set `PATH="/usr/bin:$ENV{PATH}"` and
simply call `python`. This ensures that the correct Python interpreter (from the activated virtual environment)
is used, allowing for proper discovery of installed packages.

Fixes #98

Co-authored-by: Ali Raza <araza@redhat.com>
Co-authored-by: Prarit Bhargava <prarit@redhat.com>
Signed-off-by: Emilien Macchi <emacchi@redhat.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -111,11 +111,11 @@ message("VENV_DIR ${VENV_DIR}")
 # unset(Python3_EXECUTABLE)
 # find_package(Python3 COMPONENTS Interpreter REQUIRED)
 
-execute_process(COMMAND ${CMAKE_COMMAND} -E env VIRTUAL_ENV=${VENV_DIR} "${VENV_DIR}/bin/python" -c "import site; print(site.getsitepackages()[0], end='')" OUTPUT_VARIABLE VENV_SITE)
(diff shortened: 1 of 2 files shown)


xinyazhang:
> Let me get back to you, I'm investigating this further.

If you're on windows, the executable path will be `venv/Scripts/python.exe`. In that case this change can be merged into PR that adds Windows support to AOTriton.

PR #105, opened 2025-06-27
test: use l1/l2 norm as softmax_scale

xinyazhang:
Originally we were using a fixed value 1.2 to as this scalar. However, during the integration testing we found this factor cannot detect problems with certain build options.

Now we use L1/L2 norm for this factor for better coverage.

### DIFF
--- a/tritonsrc/test_backward.py
+++ b/tritonsrc/test_backward.py
@@ -6,7 +6,6 @@ import sys
 import pytest
 import torch
 import os
-import math
 
 from _common_backward import _do_test_op_bwd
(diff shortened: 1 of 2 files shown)


xinyazhang:
TODO: confirm this can discover those known problems

tritonsrc/test_backward.py
```diff
@@ -6,6 +6,7 @@
 import pytest
 import torch
 import os
+import math
```
jeffdaily, line 9: Unused import?

jeffdaily (approved):

PR #106, opened 2025-06-30
Update README.md

xinyazhang:


README.md
```diff
@@ -87,8 +90,8 @@ upstream and may support more recent version of AOTriton
 |-----------------------|-------------------------------------------------|
 |  2.2 and earlier      |               N/A, no support                   |
 |        2.3            |                   0.4b                          |
-|        2.4            |                   0.7b (backported)             |
-|        2.5            |                   0.8b (backported)             |
+|        2.4            |                   0.9b (backported)             |
+|        2.5            |                   0.9b (backported)             |
 |        2.6            |                   0.9b (backported)             |
 |        2.7            |                   0.9b (backported)             |
 |        2.8            |                   0.10b (once released)         |
```
jithunnair-amd, line 97: Why is it that we don't list the '0.10b drop-in only" clauses here for any branch that supports 0.9b? Wouldn't the same logic apply as for upstream branches?

README.md
```diff
@@ -76,7 +76,10 @@ of AOTriton. The compatibility matrix is shown below
    suffer from runtime errors.
 3. To be specific, it is shipped with 0.9.2b. 0.9b and 0.9.1b should not be
    used in order to avoid linking issues, and confusion about version strings.
-4. 0.10b is backward compatible with 0.9b's API. However, PyTorch 2.8 will
+4. 0.10b is backward compatible with 0.9b's API. Hence it can be used as a drop-in
+   replacement for installed PyTorch wheels. However, 0.10b cannot be built with
```
jithunnair-amd, line 80: > Hence it can be used as a drop-in replacement for installed PyTorch wheels.

Can you please test this locally on your end? I wouldn't want to make that assertion without trying it at least.

README.md
xinyazhang: The same logic applies, but backport happens very frequently on ROCm forks and they will be moved to latest version at some point...

pruthvistony (approved):

PR #109, opened 2025-07-23
AITER ASM kernel Integration Phase 2: Operator Integration

xinyazhang:
WIP


jeffdaily (approved):

CMakeLists.txt
```diff
@@ -52,6 +52,7 @@ find_package(hip REQUIRED)
 
 set(VENV_DIR "${CMAKE_CURRENT_BINARY_DIR}/venv" CACHE STRING "Virtual Environment Directory")
 option(AOTRITON_NO_PYTHON "Disable python binding build" OFF)
+option(AOTRITON_USE_TORCH "Enable functions in python binding that requires libtorch" ON)
```
jammm, line 55: This could break pytorch builds that require building aotriton first before building libtorch. I think this should be OFF by default.

xinyazhang: `AOTRITON_USE_TORCH` is only applicable when `AOTRITON_NO_PYTHON=OFF`. Neither the released binary nor the PyTorch cmake file build python binding.

jammm: Ah okay, I missed that!

PR #116, opened 2025-08-17
CMakeLists.txt: AOTRITON_INHERIT_SYSTEM_SITE_TRITON flag

### LINKED ISSUE #114: [Feature]: It should be possible to reuse an existing triton
aotriton's CMakeLists.txt insists on building triton from source to create `AOTRITON_TRITON_SO`.

It would be nice to allow relying on a pre-built triton.

LunNova:
Closes #114

## Motivation

Reusing an existing triton install allows building it separately from aotriton.

This is useful for nixpkgs' packaging of aotriton as we would prefer to build triton in a separate derivation and import it in aotriton.

## Technical Details

Added a new CMake flag that's off by default that enables the `--system-site-packages` flag when creating the venv and checks an existing triton can be imported instead of building it from the submodule.

## Test Plan

Build with and without the flag enabled.

## Test Result

Works for both cases.

## Submission Checklist


LunNova:
Updated to use `AOTRITION_INHERIT_SYSTEM_SITE_TRITON`

LunNova:
Thanks!

PR #128, opened 2025-09-19
Update build instructions for 0.11

xinyazhang:
## Overview

0.11b adds LazyTensor into V3 API, which requires libTorch for python binding.
This creates a chicken-and-egg issue.

Updating build instructions with build options that requires less dependencies.


xinyazhang:
Outdated, closing.

PR #135, opened 2025-10-30
Add support of hdim_qk != hdim_vo

xinyazhang:
# Overview

This PR allows the SDPA/Flash kernel operates on inputs where Q/K have
different head dimension with V tensor.

## Note

* The code uses `hdim_vo` instead of `hdim_v` for better alignment with
  `hdim_qk`, and emphasizes the output tensor (`o`) should follow Tensor `V`'s
  head dimension.

# Major Changes

* [kernel] `head_dim` argument in all SDPA Triton kernels is replaced by
  `hdim_qk` and `hdim_vo` arguments to support this feature.
* [kernel] `composed_load` now pads all POT tensor loading to meet `hdim_qk !=
  hdim_vo` demands.
* [shim] The dispatcher will use the last dimensions of `Q` and `V` tensor as
  `hdim_qk` and `hdim_vo`
* [test] Add test cases to test `hdim_qk != hdim_vo`. New cases are added to
  + `test_fast` case (`FOR_RELEASE=0`)
  + `test_hdim_qk_ne_vo` (`FOR_RELEASE=3`). To reduce the number of testing
    cases only `seqlen <= 1024` cases are tested.
  + The algorithm to choose `hdim` pairs are documented in `test/_core_test_backward.py`
* [test] Support passing `tuple[int, int]` as `D_HEAD` to test `hdim_qk !=
  hdim_vo`

# Minor Changes

* [docs] Update comments about input dimensions in V2 header file.
* [test] Fix parsing of `torch.__version__`

# Known Problems

* Output tensor has to be created from input shapes rather than
  `torch.empty_like`, defeating the purpose to test using non-BHSD layout
  output tensor with `storage_flip`
* There may be performance impact for inputs with irregular head dimensions
  padded to NPOT head dimensions. When assuming `hdim_qk == hdim_vo`, only
  last POT sub-tensor of NPOT hdim tensor needs load with masks. Now all POT
  sub-tensors need it.

### DIFF
--- a/test/_core_test_backward.py
+++ b/test/_core_test_backward.py
@@ -320,10 +320,6 @@ def _do_test_op_bwd(request, args, device_str='cuda'):
     ref_out, _ = ctx.compute_ref_forward(sdpa_params)
 
     dout = torch.rand_like(tri_out)
-    print(f'{q.shape=}')
-    print(f'{k.shape=}')
-    print(f'{v.shape=}')
-    print(f'{tri_out.shape=}')
     if PROBE_UNSUPPORTED:
         try:
             ctx.compute_backward(tri_out, dout)
--- a/test/_common_test.py
+++ b/test/_common_test.py
@@ -424,10 +424,6 @@ class SdpaContext(object):
                                       scale=p.sm_scale,
                                       dropout_mask=dropout_mask,
                                       enable_gqa=enable_gqa)
-        print(f'{ref_q.shape=}')
-        print(f'{ref_k.shape=}')
-        print(f'{ref_v.shape=}')
-        print(f'{ref_out.shape=}')
         return (ref_out, ref_mask)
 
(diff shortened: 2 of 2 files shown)


test/_core_test_backward.py
```diff
@@ -274,6 +320,10 @@ def _do_test_op_bwd(request, args, device_str='cuda'):
     ref_out, _ = ctx.compute_ref_forward(sdpa_params)
 
     dout = torch.rand_like(tri_out)
+    print(f'{q.shape=}')
+    print(f'{k.shape=}')
+    print(f'{v.shape=}')
+    print(f'{tri_out.shape=}')
```
jeffdaily: Same here, are these for debug?  Info?  Leftover?

jeffdaily (approved):
Approved, with some nit questions about some print() statements.

PR #137, opened 2025-11-05
Adding gfx908 support

UD-mmcminn:
## Motivation

I use MI100's for AI development and it would be nice if we could keep support for those. There does not appear to be anything in aotriton that blocks MI100 support. 

## Technical Details

Just added gfx908 to the places that need it.

## Test Plan

I tried testing, but was unable to get the pytest suite to work on my test system. If you can provide some info on that I'd be happy to test.

## Test Result
 
N/A

## Submission Checklist


UD-mmcminn:
Gotcha, the issue is, this limitation is breaking pytorch for gfx908 right now. I've got a fork and a patched version of torch I am working with to try and get a test working. Having a weird error with hip and the MATH backend I am tracking down. Likely a flags error. Anyway, maybe experimentally enable it while we get it figured out and debugged?

xinyazhang:
> Anyway, maybe experimentally enable it while we get it figured out and debugged?

The problem is the capability to compile the kernel for gfx908 is questionable. IIRC there is no gfx908 MFMA support in Triton compiler, and its MFMA hardware only works with fp16 datatype.

v2python/tuning_lut.py
```diff
@@ -13,6 +13,7 @@
 SKIPPED_LUT_CHECK = os.getenv('AOTRITON_SKIP_LUT_CHECK', default='').split(',')
 
 ARCH_TO_DIRECTORY = {
+    'gfx908'  : 'amd-gfx908',
```
xinyazhang: v2python is deprecated

v2src/util.cc
```diff
@@ -46,6 +46,7 @@ struct LazyGpu {
 };
 
 std::unordered_map<std::string, GpuClassifier> LazyGpu::string_to_classifier = {
+  { "gfx908", DummyClassifier<GPU_AMD_ARCH_GFX908_MOD0 >() },
```
xinyazhang: v2src is deprecated and changes are not going to take effect.

xinyazhang (changes requested):
* Tuning database of gfx908 is missing. All kernel will be compiled with default options. (It's okay if it's intentional)
* Changes to v2src/v2python can be removed. They are deprecated.
* Need to add gfx90a to `isArchExperimentallySupported`, unless someone volunteers to run full tuning/testing process on gfx908 before each release.

UD-mmcminn:
I have generated the tuning database, well most of it. There were some configs that would cause crashes when dropout was involved. Also, larger sequence would run out of vram. The JSON is somewhat large and I am unsure what to do with it.

IMbackK:
@UD-mmcminn just curious if you got any further on this.

PR #162, opened 2026-03-23
Enable Pipelining and XCD Remapping on gfx950 GPUs

xinyazhang:
# Overview

This PR enables kernel pipelining and XCD (compute die) remapping
optimizations for gfx950 GPUs (MI350 series), significantly improving
performance of Flash Attention kernels.

Pipelining (also known as FlashAttention v3) allows overlapping memory
operations with compute operations in Triton kernels, while XCD remapping
optimizes workload distribution across multiple compute dies on CDNA
accelerators.

This update delivers 904 TFLOPS performance on MI355X with mainline Triton JIT,
improved from 753 TFLOPS with the same Triton JIT process without pipelining.

The performance is reported by
```
TRITON_PRINT_AUTOTUNING=1 USE_CAUSAL=0 N_CTX=14 D_HEAD=128 N_HEADS=64 python tritonsrc/performance_forward.py
```

## Major Changes

* [tritonsrc] Enable kernel pipelining for forward and backward kernels
  + Add pipelining configurations to autotune search space
  + Pass `num_stages` parameter to inner loop kernels when
    `MASK_STEPS=False`
  + Enable pipelining for backward kernels (dq, dk, dv)
* [tritonsrc] Add XCD remapping support for multi-die GPUs
  + Add `NUM_XCDS` kernel argument to track number of compute dies
  + Implement `remap_xcd` logic to flip `tl.program_id` 0 and 1 for
    better load balancing
  + Use SHB (sequence-head-batch) launching grid when `NUM_XCDS == 1`,
    HSB grid otherwise
* [rules] Replace `AOTRITON_ARCH_PRODUCTION_LINE` with
  `AOTRITON_ARCH_WARPSIZE`
  + Wave size and register file size matter more than chip names for
    kernel selection
  + Remove dead conditional branches
* [rules] Set `AOTRITON_PRE_LOAD_OPTIONS=2` by default
  + Pre-loading V tensor (`PRE_LOAD_V=True`) is required for optimal
    performance
* [shim] Adaptive grid calculator respects `NUM_XCDS` parameter
* [codegen] Add `PROGRAMMATIC_PERFS` to kernel descriptor
  + Allows adding performance-related fields to dataframe from
    functional objects
  + TODO: Remove deprecated `tuned_kernel$PERSISTENT_TYPE` column when
    updating database

## Minor Changes

* [tritonsrc] Refactor autotune configurations
  + Consolidate autotune and configuration generation logic
  + Search for optimal backward kernel configs (num_warps=8 tested but
    no improvement for hdim=128)
  + Move `IS_RDNA` and `NUM_XCDS` detection to `tuned_bwd.py`
* [tritonsrc] Experiment with FlashAttention v3 optimizations
* [ci] Add helper messages to `build-shim.sh` and `build-test.sh`
  + Document that `<target arch>` can be semicolon-separated list of
    architectures

# Known Issues

* From hdim=128 tests, `num_warps=8` configuration does not improve performance
  for  
  + BWD kernels
  + Navi GPUs
* Tuning database is not updated yet to utilize this change.
* `tritonsrc/test_varlen.py` is broken due to `varlen_attn_torch_function.py` is not updated


tritonsrc/tuned_bwd.py
```diff
@@ -1,152 +1,79 @@
...
+
+IS_RDNA = is_rdna()
+
+def get_num_of_xcds():
+    if evaluate_gfx_arch_within(['gfx942', 'gfx950']):
+        return 8
+    # Unknown
+    return 1
+
+NUM_XCDS = get_num_of_xcds()
```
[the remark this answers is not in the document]
xinyazhang: Reject. This project doesn't care about CPU only environment. Stop assuming this in the future.

v3src/flash/attn_bwd.cc
```diff
@@ -35,23 +35,17 @@ dim3 BwdPreprocessVarlenContext::grid_calculator() const {
...
-    AOTRITON_NS::cdiv<uint32_t>(params->max_seqlen_q, this->BLOCK_M),
-    uint32_t(params->Q->size(1)),
-    params->num_seqlens == 0 ? uint32_t(params->Q->size(0)) : std::abs(params->num_seqlens),
-  };
-  // std::cerr << "bwd_kernel_dq grid conf " << grid.x << " " << grid.y << " " << grid.z << std::endl;
-  return grid;
+  auto S = AOTRITON_NS::cdiv<uint32_t>(params->max_seqlen_q, this->BLOCK_M);
+  auto H = uint32_t(params->Q->size(1));
+  auto B = params->num_seqlens == 0 ? uint32_t(params->Q->size(0)) : std::abs(params->num_seqlens);
+  return NUM_XCDS > 1 ? dim3 { H, S, B } : dim3 { S, H, B };
```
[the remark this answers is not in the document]
xinyazhang: Reject. `this->` was the legacy practice for templated functions. Unnecessary for v3 dispatcher.

v3python/rules/flash/_common.py
```diff
@@ -50,6 +50,10 @@ class FlashKernel(KernelDescription):
     LUT_FULL_SEQLEN_K = [16,32,64,128,256,512,1024,2048,4096,8192]
     LUT_FULL_SEQLEN_NAVI = [16,32,64,128,256,512,1024]
 
+    PROGRAMMATIC_PERFS = {
+        "NUM_XCDS": lambda f: 8 if f.arch in ['gfx942', 'gfx950'] else 1
```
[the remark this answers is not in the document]
xinyazhang: Reject. attn_fwd.PERF_CHOICES are only default values.

tritonsrc/bwd_inner_dk_dv.py
```diff
@@ -88,7 +91,7 @@ def bwd_inner_dk_dv(
     start_k: select k and dV
     start_q: select q and dO
     '''
-    for block_index in range(nblocks_1+nblocks_2):
+    for block_index in range(nblocks_1+nblocks_2, num_stages=NUM_STAGES):
```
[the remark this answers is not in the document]
xinyazhang: Reject. This is Triton kernel. Stop assume it's python code

tritonsrc/bwd_inner_dq.py
```diff
@@ -82,7 +85,7 @@ def bwd_inner_dq(
     dV: (seqlen_k, hdim)
     '''
     # for start_k in range(lo, hi, BLOCK_N):
-    for block_index in range(nblocks_1+nblocks_2):
+    for block_index in range(nblocks_1+nblocks_2, num_stages=NUM_STAGES):
```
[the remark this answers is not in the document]
xinyazhang: Not python code

tritonsrc/fwd_kernel.py
```diff
@@ -107,7 +138,32 @@ def attn_fwd(
...
+    tl.assume(stride_kn >= 0)
+    tl.assume(stride_kk == 1)
+    tl.assume(stride_vz >= 0)
+    tl.assume(stride_vh >= 0)
+    tl.assume(stride_vk >= 0)
+    tl.assume(stride_vn == 1)
+    tl.assume(stride_oz >= 0)
+    tl.assume(stride_oh >= 0)
+    tl.assume(stride_om >= 0)
+    tl.assume(stride_on == 1)
```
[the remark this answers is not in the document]
xinyazhang: Are you sure tl.device_assert will provide the same compiler hints as tl.assume?

tritonsrc/bwd_inner_dk_dv.py
jeffdaily, line 94: **Bug: `range()` should be `tl.range()`, and the `NM_STAGES` condition is inverted**

Two issues here:
1. **Wrong function**: `range(nblocks_1+nblocks_2, um_stages=NUM_STAGES)` should be `tl.range(...)`. Plin Python `range()` does not accept a `num_stages` kyword argument — this will fail at compile time on the backward pat.
2. **Inverted condition**: `NUM_STAGES = None if ULL_BLOCKS else 1` sets `NUM_STAGES=None` precisely hen `FULL_BLOCKS=True`, which is the common hot path(non-causal, non-masked). Pipelining with `num_stage=4` should be enabled when `FULL_BLOCKS=True`, not dsabled. The same issue exists in `bwd_inner_dq.py` ad `bwd_inner_fuse.py`.
Suggested fix:
```python
NUMSTAGES: tl.constexpr = 4 if FULL_BLOCKS else 1
for bock_index in tl.range(nblocks_1+nblocks_2, num_stage=NUM_STAGES):
```

*(Review by Claude Sonnet 4.6)*

tritonsrc/tuned_bwd.py
```diff
@@ -1,152 +1,77 @@
...
+import torch
 
 from flash import (
     bwd_preprocess as bare_bwd_preprocess,
     bwd_kernel_dk_dv as bare_bwd_kernel_dk_dv,
     bwd_kernel_dq as bare_bwd_kernel_dq,
 )
 
+def evaluate_gfx_arch_within(arch_list):
+    assert torch.cuda.is_available()
```
jeffdaily, line 12: **CUDA availability guard removed — will crash on CU-only nodes at import time**

The old `evaluate_gfxarch_within` in `attn_torch_function.py` guarded wit `if not torch.cuda.is_available(): return False`. Tis new version replaces that with `assert torch.cuda.is_available()`, which will raie an `AssertionError` at module import time on any CPU-only machine (e.g. CI uild nodes, documentation builders).

Suggested fix:
```python
def evaluate_gfx_arch_within(arch_list):
    if not torch.cuda.is_available():
        return False
    gcn_arch_name = torch.cuda.get_device_properties'cuda').gcnArchName
    return any(arch in gcn_arch_name for arch in arh_list)
```

*(Review by Claude Sonnet 4.6)*

tritonsrc/attn_torch_function.py
```diff
@@ -802,70 +586,8 @@ def backward_split(ctx, do, _, fwd_tuning_result):
...
-                    }
-                    tuning_result = {
-                        'kernel_name' : 'bwd_kernel_dk_dv',
-                        'inputs' : inputs,
-                        'tuned_kernel' : tuned_kernel,
-                        'compiler_options' : compiler_options,
-                    }
-                    """
                     ctx.tuning_result.append(('bwd_kernel_dk_dv', tuning_result))
                     print(f'{id(ctx.tuning_result)=}')
```
jeffdaily: **Debug print statement left in — will fire on every backward pass**

`print(f'{id(ctx.tuning_result)=}')` is a debug print that will emit on every training step during the backward pass. This (and a second identical occurrence later in `backward_split`) should be removed before merge. The `# print(f'backward ...')` earlier in this function was correctly commented out — these need the same treatment.

*(Review by Claude Sonnet 4.6)*

tritonsrc/varlen_attn_torch_function.py
```diff
@@ -340,8 +340,8 @@ def forward(ctx, q, k, v, seqlen_q, seqlen_k, causal, sm_scale, dropout_p,
     def backward(ctx, do, _, fwd_tuning_result):
         q, k, v, b, o, L = ctx.saved_tensors
         # if q.shape[-1] <= 32:
-        Lq, Lk, Lv = q.shape[-1], k.shape[-1], v.shape[-1]
-        assert Lq == Lk and Lk == Lv and Lk == ctx.head_dim
+        Lq, Lk, Lv, Lo = q.shape[-1], k.shape[-1], v.shape[-1], v.shape[-1]
+        assert Lq == Lk and Lv == Lo
```
jeffdaily: **Bug: `Lo` comparison is a self-comparison — silenly drops an output shape check**

On line 343 (new),`Lo` is assigned as `v.shape[-1]`, the same value as`Lv`. So `assert Lv == Lo` is comparing `v.shape[-1] to itself, which is always `True` and provides no sfety guarantee.

The original intent was presumably o check the output tensor `o`'s head dimension (i.e. `Lo = o.shape[-1]`). The same bug appears in `attn_orch_function.py` in the `backward_split` method.

Sggested fix (line 343):
```python
Lq, Lk, Lv, Lo = qshape[-1], k.shape[-1], v.shape[-1], o.shape[-1]
assrt Lq == Lk and Lv == Lo
```

*(Review by Claude Sonet 4.6)*

tritonsrc/fwd_kernel.py
```diff
@@ -48,6 +48,37 @@
 def cdiv_fn(x, y):
     return (x + y - 1) // y
 
+@triton.jit
+def remap_xcd(pid, GRID_MN, NUM_XCDS: tl.constexpr = 8):
```
jeffdaily, line 52: **Minor: `remap_xcd` call sites are inconsistent inhow they pass `NUM_XCDS`**

The three call sites pas `NUM_XCDS` differently:
- `fwd_kernel.py`: `remap_xd(tl.program_id(0), Num_head_q, NUM_XCDS=NUM_XCDS)` keyword)
- - `bwd_kernel_dq.py`: `remap_xcd(tl.programid(0), num_head_q, NUM_XCDS=NUM_XCDS)` (keyword)
- - - `wd_kernel_dk_dv.py`: `remap_xcd(tl.program_id(0), nu_head_k, NUM_XCDS)` (positional)
All three are funcionally correct since `NUM_XCDS` is the third parameer, but please standardize on the keyword form for carity and to avoid future argument-order confusion.
*(Review by Claude Sonnet 4.6)*

pruthvistony (approved):

jeffdaily (changes requested):
**Code review by Claude Sonnet 4.6**

This PR delivers meaningful performance improvements for gfx950/MI350X via XCD remapping and forward kernel pipelining. The `remap_xcd` logic is well-designed and well-commented, the autotune boilerplate elimination is a nice cleanup, and the stride `tl.assume` hints are a good addition. However, there are several correctness issues that should be addressed before merge:

**Must fix:**
1. **Backward pipelining is broken**: `bwd_inner_dk_dv.py`, `bwd_inner_dq.py`, and `bwd_inner_fuse.py` all use `range(..., num_stages=...)` instead of `tl.range(...)`. Plain `range()` does not accept `num_stages` — this will fail at compile time on the backward hot path. Additionally the `NUM_STAGES` condition is inverted (`None if FULL_BLOCKS` should be the opposite of the intent).
2. 2. **CUDA guard regression in `tuned_bwd.py`**: `assert torch.cuda.is_available()` replaces the previous `if not ...: return False` guard, causing an `AssertionError` at import time on CPU-only nodes (CI build servers, doc builders, etc.).
3. 3. **`Lo` self-comparison in `varlen_attn_torch_function.py` and `attn_torch_function.py`**: `Lo = v.shape[-1]` means `assert Lv == Lo` is a no-op that silently drops an output dimension check.
**Should fix:**
4. Two `print(f'{id(ctx.tuning_result)=}')` debug statements remain in `attn_torch_function.py` and will fire on every backward pass.
5. 5. `remap_xcd` call in `bwd_kernel_dk_dv.py` uses positional arg for `NUM_XCDS` while the other two call sites use keyword arg — minor inconsistency worth standardizing.
The forward path (XCD remapping + `tl.range` pipelining) looks solid. The blocking issues are all in the backward path and are straightforward to fix.

tritonsrc/bwd_inner_dk_dv.py
xinyazhang: Apparently AI Agent doesn't understand Triton, Copilot also complains but this code is complete legit for Triton and I will not change right now it because tl.range may have unforeseeable consequences.

tritonsrc/tuned_bwd.py
xinyazhang: Copilot also complains, but I will not fix it. Supporting CPU only environment is completely pointless for code under `tritonsrc` (Triton kernel and related Triton-only tests).

tritonsrc/attn_torch_function.py
xinyazhang: Fixed

tritonsrc/varlen_attn_torch_function.py
xinyazhang: Fixed

tritonsrc/fwd_kernel.py
xinyazhang: Fixed

tritonsrc/bwd_inner_dk_dv.py
xinyazhang: I've added `.github/instructions/code-review.instructions.md` to disable this check in Triton kernel code.

tritonsrc/tuned_bwd.py
xinyazhang: I've added `.github/instructions/code-review.instructions.md` to let copilot assume GPU environment in this directory.

tritonsrc/varlen_attn_torch_function.py
```diff
@@ -340,8 +340,8 @@ def forward(ctx, q, k, v, seqlen_q, seqlen_k, causal, sm_scale, dropout_p,
     def backward(ctx, do, _, fwd_tuning_result):
         q, k, v, b, o, L = ctx.saved_tensors
         # if q.shape[-1] <= 32:
-        Lq, Lk, Lv = q.shape[-1], k.shape[-1], v.shape[-1]
-        assert Lq == Lk and Lk == Lv and Lk == ctx.head_dim
+        Lq, Lk, Lv, Lo = q.shape[-1], k.shape[-1], v.shape[-1], o.shape[-1]
+        assert Lq == Lk and Lv == Lo
         head_dim_factors = factor_head_dim(Lk)
         head_dim_rounded = sum(head_dim_factors)
         padded_head = head_dim_rounded != ctx.head_dim
```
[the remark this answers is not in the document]
xinyazhang: `padded_head = head_dim_rounded != max_head_dim` should be `padded_head = head_dim_rounded != min(Lk, Lv)`

PR #167, opened 2026-04-20
Remove unnecessary Windows ccache guard

astrelsky:
## Motivation

Enable use of ccache on Windows.

## Technical Details

Building the aotriton runtime on Windows behaves the same with and without ccache. The problem the removed lines intended to guard affects compiling even without ccache and the solution is to use `chcp 65001` to force utf-8 mode. While how that gets documented is open to debate, I won't include it in this pull request and am simply removing the guard that doesn't have the intended effect and will hinder building the aotriton runtime with ccache when building pytorch.

## Test Plan

Tested locally.

## Test Result

Built successfully. These stats are for the entire `pytorch`, `torchaudio` and `torchvision` builds.
```
Cacheable calls:   6745 /  6764 (99.72%)
  Hits:            6462 /  6745 (95.80%)
    Direct:        6462 /  6462 (100.0%)
    Preprocessed:     0 /  6462 ( 0.00%)
  Misses:           283 /  6745 ( 4.20%)
Uncacheable calls:   19 /  6764 ( 0.28%)
Local storage:
  Cache size (GB):  8.9 / 100.0 ( 8.93%)
  Hits:            6462 /  6745 (95.80%)
  Misses:           283 /  6745 ( 4.20%)
```

## Submission Checklist


astrelsky:
@amd-nicknick just for tracking purposes. I suspect any change in aotriton would take time to make it's way into pytorch.

xinyazhang:
@jammm can you confirm if more recent `ccache` works on windows?

jammm:
I don't use ccache on my local builds. @ScottTodd does the latest ccache on Windows support multi-byte characters now?

astrelsky:
> I don't use ccache on my local builds. @ScottTodd does the latest ccache on Windows support multi-byte characters now?

I have already built aotriton and the aotriton images with cache enabled on Windows without a problem multiple times using `chcp 65001`. I don't think the problem ever was a ccache issue since I encountered the problems without ccache and having the default console code page (not setting `chcp 65001`). I am currently using the latest version of cache though so if ccache fixed something, then it's fixed in cache and this should go away.

jammm:
> > I don't use ccache on my local builds. @ScottTodd does the latest ccache on Windows support multi-byte characters now?
> 
> I have already built aotriton and the aotriton images with cache enabled on Windows without a problem multiple times using `chcp 65001`. I don't think the problem ever was a ccache issue since I encountered the problems without ccache and having the default console code page (not setting `chcp 65001`). I am currently using the latest version of cache though so if ccache fixed something, then it's fixed in cache and this should go away.

Nice. Scott did file an issue about the multi-byte character support a while ago and it was fixed in https://github.com/ccache/ccache/commit/441da4b4408e5e21a15e5f3c6cc6e50a5db92842. The fix has been available since v4.12.

Can you enforce a version of 4.12 here, so we know the fix is in place? In case 4.12 isn't detected, let it fallback to the non-ccache path like before.

astrelsky:
> > > I don't use ccache on my local builds. @ScottTodd does the latest ccache on Windows support multi-byte characters now?
> > 
> > I have already built aotriton and the aotriton images with cache enabled on Windows without a problem multiple times using `chcp 65001`. I don't think the problem ever was a ccache issue since I encountered the problems without ccache and having the default console code page (not setting `chcp 65001`). I am currently using the latest version of cache though so if ccache fixed something, then it's fixed in cache and this should go away.
> 
> Nice. Scott did file an issue about the multi-byte character support a while ago and it was fixed in https://github.com/ccache/ccache/commit/441da4b4408e5e21a15e5f3c6cc6e50a5db92842. The fix has been available since v4.12.
> 
> Can you enforce a version of 4.12 here, so we know the fix is in place? In case 4.12 isn't detected, let it fallback to the non-ccache path like before.

Yes I can add it tonight. Any suggestions on the best approach for checking the version in cmake?

jammm:
I believe `find_package()` has an argument to specify version number. I'd try that and then attempt a build with ccache 4.12 and, say, ccache 4.11 to test if it works. I'm not a cmake expert myself, though alternatively, you can play around with logging using `[message](https://cmake.org/cmake/help/latest/command/message.html)("...")` to log the variables set by `find_package` after it finds 4.12+ ccache.

astrelsky:
I set it to check for `4.13` as a test and moved the ccache that comes with strawberry perl 🤮 higher in the path. So I know it is working correctly now. 
```
-- skipping ccache version 4.12.1

-- using ccache version 4.13.4

-- [AOTriton] ccache found and enabled
-- [AOTriton] Triton's caching system is parallel to ccache and always enabled.
-- [AOTriton] UTF-8 input character is set for all C/C++ source files.
```

xinyazhang:
@astrelsky @jammm any updates? If no object I'll merge this as part of 0.12 release

jammm (approved):

astrelsky:
> @astrelsky @jammm any updates? If no object I'll merge this as part of 0.12 release

I didn't notice the ping for me this morning, I only saw jammm mentioned. No objections here, sorry for the late response 😅.

PR #170, opened 2026-05-01
Tuner V3.5 and Tuning Database.

xinyazhang:
# Overview

This is a jumbo PR that delivers Tuner v3.5: a complete rewrite of the AOTriton
tuning infrastructure replacing Celery/RabbitMQ/Ray with a simpler, more
debuggable stack (PostgreSQL queue + Unix socket local broker), plus a
browser-based WebUI for the full tuning lifecycle.

# Critical Bug Fix

## GQA + Bias Backward Kernel Correctness (`tritonsrc/`)

`bwd_kernel_dk_dv` and the dk/dv branch of `bwd_kernel_fuse` computed `B_ptr`
using `off_h_k` (K-head index) **before** the Q-head inner loop. Under GQA
(multiple Q-heads per K-head), every Q-head in the group read the same bias
row instead of its own, producing silently wrong gradients for any model that
combines GQA with attention bias.

Fix: move `B_ptr` initialization inside the `for off_h_q` loop, indexing by
`off_h_q` to match the invariant that bias is always Q-head indexed. The
now-incorrect "CAVEAT: bias is incompatible with GQA" comments are removed
along with the bug.

Affected files: `tritonsrc/bwd_kernel_dk_dv.py`, `tritonsrc/bwd_kernel_fuse.py`

# Major Changes

## Tuner v3.5 Core Infrastructure

* [localq] Replace Ray ActorPool with Unix socket-based `LocalBroker`
  + One broker process per node routes work to GPU workers via `/wkdir/run/broker.sock`
  + DAG workflow: preprocess → probe → tune_hsaco (parallel round-robin) → postprocess
  + Application-level JSON-lines logging; no special tools needed to debug hangs
  + Remove Ray, RabbitMQ, and Celery dependencies
* [pq] PostgreSQL replaces RabbitMQ as the task queue
  + Partitioned `task_queue` table (one partition per arch)
  + `tuning_results`, `best_tuning_results`, `task_extra_uts` tables
  + `TaskQueue`, `ReconnectableConn`, and result helpers in `v3python/tune/pq/`
  + `task_extra_uts.im_text` enforces no-space via CHECK constraint
* [worker] Each GPU node runs: 1 broker, N GPU workers, 4 PG readers, 4 CPU workers
  + Signal-based graceful shutdown; zombie reaping; epoll-based buffered I/O
  + Exaid timeout set to 30 s; VRAM clamping ported from tuner v2

## WebUI

* [webui] Full browser-based management interface (Flask + HTMX, mTLS)
  + **Workers tab**: register hosts, detect GPU metadata, GPU selection for
    start/restart, office hours scheduler, worker status with container ID and
    GPU process counts
  + **Builds tab**: remote build node config, library build, test library build,
    workdir sync, Docker image build — all with live streaming command output
  + **Servers tab**: PostgreSQL start/stop/status, schema init/recreate,
    compute_best_results, export_best_results, bake LUT (populate MV, sancheck,
    decompose)
  + **Testing tab**: per-host run-test (full / partial / record-adiffs), tail
    output, show failures window, download adiffs file
  + **Dashboard**: tuning progress table with per-arch pending/running/completed,
    ETA, stale task count; auto-refresh
  + **Debug tab**: inspect `task_queue` entries and all related rows

## Testing Infrastructure

* [ci] `run-test.sh` writes output to `$OUTPUT_DIR`; `PARTIAL_INFO_DIR` drives
  selective re-runs via `--select-from-file` on pytest
* [single] `run-test.sh` converted to named arguments (`--pass`, `--hostname`, …)
* [webui/testing] `variant=partial` runs tests against failing subset;
  `variant=partial_adiffs` additionally sets `RECORD_ADIFFS_TO` for accuracy-diff capture
* [bin] `amend_sel_to_adiffs.sh` creates adiffs.txt from sel\<pass\>.txt

## Database Pipeline

* [compute_best_results] `--incremental` and `--fix [host:]<pass>` modes;
  `ACCURACY_MULTIPLIER` tightened to 3×; fixes `cd` to repo root before `python -m`
* [export_best_results] Builds centralized tuning SQLite from PG best results
* [bake_lut] Full pipeline: populate most-accurate MV → compute best → sancheck → decompose → export
* [retry/reset] `reset_broken_to_pending` collects UT-failed entries and propagates
  extra UTs end-to-end through preprocessing; `retry_missing_entries` resets
  missing LUT entries to pending

## Flash Attention

* [rules/database] Add gfx950 (MI350) tuning database and FAv3 kernel support
* [rules/database] Add gfx1100 and gfx1201 tuning databases
* [module] `FlashEntry.as_text()` recursive formatter never emits spaces in
  tuples/lists (required for `task_extra_uts.im_text` constraint)
* [rules] `attn_fwd` gfx950: dedicated config block with `BLOCK_M=256, BLOCK_N=64,
  num_warps=8, num_stages=4`

# Minor Changes

* [lib] `sqlite3_compat.sh` shim for containers without SQLite CLI
* [pq] `ReconnectableConn` auto-reconnects on dropped DB connections
* [bin] `psql` wrapper supports `-c` for non-interactive use; `pg_dump` adds zstd
* [webui] Guest read-only dashboard mode; minimize button on command outputs;
  auto-minimize on successful commands
* [CLAUDE.md] Document PostgreSQL connection pattern, `pq` package conventions,
  `config.rc` usage, argparse underscore convention

# Known Issues

* SLURM integration (design complete, implementation pending)
* WebUI-based task dispatch dispatches all combinations; fine-grained filtering
  requires CLI `dispatch` script


jammm:
> [rules/database] Add gfx1100 and gfx1201 tuning databases

Would be good to have a tuning database for gfx1151 too.
BTW, what happens on archs other than the above ? does it still use FA?

xinyazhang:
> Would be good to have a tuning database for gfx1151 too.

It's not very practical for current tuning mechanism. We need 8 GPUs at least to complete the tuning at reasonable amount of time.

To reduce the tuning workload we need to model the compiler behavior with machine learning, which is not a straightforward work.

> BTW, what happens on archs other than the above ? does it still use FA?

The target GPUs are determined by the default in CMakeLists.txt
For the rest pytorch falls back to math backend.

mgehre-amd:
> Would be good to have a tuning database for gfx1151 too. BTW, what happens on archs other than the above ? does it still use FA?

I might have some gfx1151 to spent. I'm currently running the "split" l2 validation on gfx1151, it it looks like it will take 30h on a single GPU. How long do we expect the tuning to take? I think we could ignore the backward pass for gfx1151.

xinyazhang:
> > Would be good to have a tuning database for gfx1151 too. BTW, what happens on archs other than the above ? does it still use FA?
> 
> I might have some gfx1151 to spent. I'm currently running the "split" l2 validation on gfx1151, it it looks like it will take 30h on a single GPU. How long do we expect the tuning to take? I think we could ignore the backward pass for gfx1151.

The key part is to use mGPU for testing, which has become possible with pytest-xdist but needs careful design.
Tuning on gfx115x is not possible yet due to the limits of available GPUs.

mgehre-amd:
I validated this PR on gfx1151, level 2, `split`. I have run all tests split into 512 chunks. Results are
```
layout                   done/N     wall  |  back pass  fail  skip |   var pass  fail  skip
-----------------------------------------------------------------------------------------------
split l2 n512          512/512     29.3h  |     302308     0 106936 |       1800     0     0
```

xinyazhang:
> I validated this PR on gfx1151, level 2, `split`. I have run all tests split into 512 chunks. Results are
> 
> ```
> layout                   done/N     wall  |  back pass  fail  skip |   var pass  fail  skip
> -----------------------------------------------------------------------------------------------
> split l2 n512          512/512     29.3h  |     302308     0 106936 |       1800     0     0
> ```

@mgehre-amd thanks for running the tests. I'll move gfx1151 out of experimental in another PR.

PR #179, opened 2026-06-01
fix(gfx1151): explicit out_dtype=tl.float32 in fwd tl.dot to prevent bf16/fp16 NaN

### LINKED ISSUE #54: [Issue]: large bf16 inputs leads to nan
### Problem Description

This is due to the core logic of FA algorithm:
https://github.com/ROCm/aotriton/blob/f6b28a9b7265b69e3df54ea6ba0237e8a8d6f736/tritonsrc/fwd_kernel_inner.py#L95-L97

`qk_scale * tl.max(qk, 1))` uses `vmul`, but `qk * qk_scale - m_ij[:, None]` uses `vfma`
`vfma` has higher precision than `vmul`+`vadd` because it only round once.

More specifically, Suppose the vectors all only contain one element, and `qk = round(133120.0 * 133120.0) = 17720934400.0`, `qk_scale = round(0.25 * 1.44269502162933349609) = 0.360673755407333374023`
Then
```
qk * qk_scale - m_ij
= qk * qk_scale - qk_scale * qk
= fma(qk, qk_scale, -vmul(qk, qk_scale))
= fma(qk, qk_scale, -round(17720934400.0, 0.360673755407333374023))
= round(qk * qk_scale - 6391475712.0)
= round(17720934400.0 * 0.360673755407333374023 - round(17720934400.0, 0.360673755407333374023))
= round(6391475959.3749999999922470912 - 6391475712.0)
= 247.375
```
Therefore, `p` in the code above yields to `exp2(247.375)=inf`. This consequently leads to nan in following steps.

### Operating System

N/A

### CPU

N/A

### GPU

MI300X

### ROCm Version

ROCm 6.2.3

### ROCm Component

_No response_

### Steps to Reproduce

_No response_

### (Optional for Linux users) Output of /opt/rocm/bin/rocminfo --support

_No response_

### Additional Information

_No response_

marcusr-amd:
# Overview

This PR fixes a NaN regression on gfx1151 (Strix Halo / RDNA 3.5) in the forward attention kernel under bf16 and fp16.

The packaged AOT binary produces NaN at the SAM2.1 Hiera-Large global-attention shape (`B=4, H=8, N=4096, D=72`) — 59.4% of images under bf16, 7.5% under fp16, measured on n=2693 COCO val2017 images. The JIT path from the same source is clean, indicating an AOT-vs-JIT divergence in the compiled kernel.

We believe the AOT compiler for gfx1151 may select a lower-precision accumulator for `tl.dot` at certain `BLOCK_M × BLOCK_N` configurations when `out_dtype` is not explicitly specified. Adding `out_dtype=tl.float32` makes the fp32 accumulation guarantee explicit, matching the canonical FlashAttention-2 pattern.

Fixes #54

## Major Changes

* [kernel] Add `out_dtype=tl.float32` to all three `tl.dot` calls in the non-INT8 forward path of `fwd_kernel_inner.py` (lines 170–172).
  + Prevents the Triton compiler from keeping the dot result in the input dtype (bf16/fp16) before the `Qk_scale` multiply.
  + On gfx1151, intermediate values that exceed `fp16_max` (65504) overflow to `+inf` before scaling, causing `inf − inf = NaN` in the centering step.
* [test] Add `test_issue54_gfx1151_bf16_nan.py`: 17 targeted regression tests covering the production shape with adversarial (fp16 overflow), random (4 seeds × {bf16, fp16}), accuracy, and large-sigma (σ=5/10/15) cases.
  + Tolerances match the seqlen-scaled formula used by `test_forward.py`.
  + Reference uses `torch.ops.aten._scaled_dot_product_attention_math` (same as the existing test suite).

## Minor Changes

* [kernel] Add TODO comment on the `INT8_GEMM` path in `fwd_kernel_inner.py` (line 159) — same `tl.dot` pattern without `out_dtype`, untested on gfx1151.
* [kernel] Add TODO comments in `bwd_inner_dk_dv.py`, `bwd_inner_dq.py`, and `bwd_inner_fuse.py` `dot()` helpers — same pattern, pending a backward NaN reproducer on gfx1151.

## Known Issues

* The backward `dot()` helper in `bwd_inner_{dk_dv,dq,fuse}.py` has the same `tl.dot` without `out_dtype`. A backward NaN reproducer on gfx1151 is needed before applying the fix there.
* Tests use `autotune=False` (matching `test_forward.py` convention), so they exercise the default BLOCK config, not the specific autotuned config that may trigger the bug in the AOT binary.
* The adversarial overflow test covers fp16 only. The bf16 NaN mechanism (59% rate in production) is different — bf16 shares fp32's exponent range so C²=67600 does not overflow. The bf16 NaN is believed to be a mantissa quantization issue exercised by the random-seed and large-sigma tests.

## Verification

Tested on gfx1151 (Strix Halo, Ryzen AI Max 395), JIT path:

```
TRITON_F32_DEFAULT=ieee python -m pytest tritonsrc/test_issue54_gfx1151_bf16_nan.py -v
# 17 passed in 2.6s
```

Also verified with the existing `test_forward.py` hdim72 bf16 suite:

```
TRITON_F32_DEFAULT=ieee python -m pytest tritonsrc/test_forward.py -k "hdim72 and dtype1 and CausalOff and not dropout" -v
# 2048 passed in 163s
```

Production NaN rates (measured before fix, packaged AOT binary, `rocm7.14.0a20260526`, `TORCH_ROCM_AOTRITON_ENABLE_EXPERIMENTAL=1`):

| dtype | NaN rate (n=2693 COCO val images) |
|-------|-----------------------------------|
| bf16  | 59.4%                             |
| fp16  |  7.5%                             |
| fp32  |  0%                               |

### DIFF
--- a/tritonsrc/bwd_inner_dk_dv.py
+++ b/tritonsrc/bwd_inner_dk_dv.py
@@ -22,6 +22,10 @@ def dot(BLOCK_M : tl.constexpr, QDIM : tl.constexpr, KDIM : tl.constexpr, q, k):
     if BLOCK_M == 1:
         return tl.sum(tl.view(q, [QDIM]) * tl.view(k, [KDIM]))
     else:
+        # TODO: add out_dtype=tl.float32 here once a backward NaN reproducer on
+        # gfx1151 is available to validate the fix.  The forward path already
+        # carries explicit out_dtype=tl.float32 (see issue #54); the backward
+        # dot() helper is suspected to have the same AOT-vs-JIT divergence.
         return tl.dot(q, k)
 
 @triton.jit
--- a/tritonsrc/bwd_inner_dk_dv.py
+++ b/tritonsrc/bwd_inner_dk_dv.py
@@ -22,10 +22,7 @@ def dot(BLOCK_M : tl.constexpr, QDIM : tl.constexpr, KDIM : tl.constexpr, q, k):
     if BLOCK_M == 1:
         return tl.sum(tl.view(q, [QDIM]) * tl.view(k, [KDIM]))
     else:
-        # TODO: add out_dtype=tl.float32 here once a backward NaN reproducer on
-        # gfx1151 is available to validate the fix.  The forward path already
-        # carries explicit out_dtype=tl.float32 (see issue #54); the backward
-        # dot() helper is suspected to have the same AOT-vs-JIT divergence.
+        # TODO(#54): add out_dtype=tl.float32 (see PR #179)
         return tl.dot(q, k)
 
 @triton.jit
--- a/tritonsrc/bwd_inner_dq.py
+++ b/tritonsrc/bwd_inner_dq.py
@@ -22,8 +22,7 @@ def dot(BLOCK_M : tl.constexpr, QDIM : tl.constexpr, KDIM : tl.constexpr, q, k):
     if BLOCK_M == 1:
         return tl.sum(tl.view(q, [QDIM]) * tl.view(k, [KDIM]))
     else:
-        # TODO: add out_dtype=tl.float32 once a backward NaN reproducer on
-        # gfx1151 is available (see issue #54 and fwd_kernel_inner.py fix).
+        # TODO(#54): add out_dtype=tl.float32 (see PR #179)
         return tl.dot(q, k)
 
 @triton.jit
--- a/tritonsrc/bwd_inner_dq.py
+++ b/tritonsrc/bwd_inner_dq.py
@@ -22,6 +22,8 @@ def dot(BLOCK_M : tl.constexpr, QDIM : tl.constexpr, KDIM : tl.constexpr, q, k):
     if BLOCK_M == 1:
         return tl.sum(tl.view(q, [QDIM]) * tl.view(k, [KDIM]))
     else:
+        # TODO: add out_dtype=tl.float32 once a backward NaN reproducer on
+        # gfx1151 is available (see issue #54 and fwd_kernel_inner.py fix).
         return tl.dot(q, k)
 
 @triton.jit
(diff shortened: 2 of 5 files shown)


xinyazhang (changes requested):
LGTM, but please make the the TODO line more compact.
You can leave the github PR link there.

xinyazhang:
> LGTM, but please make the the TODO line more compact. You can leave the github PR link there.

btw, confirmed `tl.dot(q0, k0, out_dtype=tl.float32)` does not cause performance regression on MI300X.
I'm speculating Triton on gfx1151 does not fuse, or failed to fuse `qk += scale * tl.dot(q, k)` into F32 WMMA instruction.

marcusr-amd:
I asked an agent to investigate whether there was any open issues that fit your speculation and this is what it came up with.

> This appears to be the same LLVM MachineSink bug as ROCm/triton#909 (SWDEV-570875) — the V_WMMA_F32_16X16X16 instruction is incorrectly sunk on gfx11xx under certain argument attribute configurations that differ between JIT and AOT paths.

  - The speculation is correct — the issue is NOT in the Triton frontend (out_dtype already defaults to tl.float32; our explicit annotation is a no-op in the IR)
  - The bug is in the LLVM MachineSink pass for gfx11xx, which incorrectly sinks the V_WMMA_F32_16X16X16_{F16,BF16} instruction
  - This is tracked as SWDEV-570875 (internal AMD LLVM bug), surfaced via ROCm/triton#909 on gfx1100
  - Our gfx1151 NaN is the same bug on a different RDNA 3.x chip
  - The upstream Triton fix (PR #8901, disable maybeDeduplicate for AMDWmma) was closed without merging — the issue is still open
  - DISABLE_LLVM_OPT="disable-machine-sink" confirms the root cause
  - Our out_dtype=tl.float32 fix may help indirectly (by changing the IR shape enough to avoid the sinking pattern) but doesn't address the underlying LLVM bug

xinyazhang (changes requested):
Refer to #179 (this PR) directly. #54 has different root cause, which happens on all architectures (including CPUs).

marcusr-amd:
Further investigation suggests that out_dtype=tl.float32 should be the default so this should be a noop.  Withdrawing until this is fully understood.

PR #184, opened 2026-06-06
Support Windows wide paths for packed kernels

jammm:
## Summary

- Use native UTF-16 Windows paths for packed-kernel runtime image loading.
- Locate `aotriton.images` on Windows with `GetModuleFileNameW` instead of the POSIX `dladdr` path.
- Pass `std::filesystem::path` through to `CreateFileW` without a UTF-8 round trip.
- Sanitize worker shard directory names derived from glob selectors like `flash/affine/*` so Windows configure/build does not try to create a directory containing `*`.

## Notes

The shard-path sanitization only affects local `Bare.shards` worker directories.

It does **not** rename:
- SHA256 image filenames
- ZIP filenames
- ZIP entries
- AKS2 entries
- generated runtime package paths

Linux remains on the existing `dladdr` / `open` / `openat` path behind the non-`_WIN32` branch.

## Validation

- Built runtime-only/noimage on Windows with `CC` and `CXX` set to `_rocm_sdk_devel\lib\llvm\bin\clang-cl.exe`.
- Passed normal installed-runtime smoke test:

  `test/test_forward.py::test_op_fwd[False-0.0-dtype0-0.0-CausalOff-4-4-16-1-1]`

- Passed the same `attn_fwd` smoke test from an install path containing a non-BMP character under `cmd.exe` with `chcp 437`.

  The test process reported:

  `console_cp 437 console_output_cp 437`

- Linux sanity checked with `_WIN32` unset: `v3src/packed_kernel_common.cc` compiles as C++20 in WSL Ubuntu, confirming the non-Windows branch is still valid.


xinyazhang (changes requested):
Overall looks good. There is one design change needed: Let's use `fd_t` to for file descriptors, it should be `int` on POSIX and `HANDLE` on Win32. In addition, let's call Win32 API directly rather than using MSVCRT's simulated C API.

xinyazhang:
Also we can remove dependency of `dlfcn-win32` from build system and README.md.

v3src/packed_kernel_common.cc
```diff
@@ -32,15 +36,54 @@ constexpr int AOTRITON_LZMA_BUFSIZ = 64 * 1024;
...
+  HMODULE module = nullptr;
+  if (!GetModuleHandleExW(GET_MODULE_HANDLE_EX_FLAG_FROM_ADDRESS |
+                          GET_MODULE_HANDLE_EX_FLAG_UNCHANGED_REFCOUNT,
+                          reinterpret_cast<LPCWSTR>(address),
+                          &module)) {
+    return {};
+  }
+
+  std::wstring path(MAX_PATH, L'\0');
+  while (true) {
```
xinyazhang: Another thing, is indefinite loop really what we want?

jammm:
Probably not, I’ll check tomorrow. Re. The dlfcn-win32 dependency, is it
not needed anymore ? I thought we still needed to dlopen.

xinyazhang:
> Re. The dlfcn-win32 dependency, is it not needed anymore ? I thought we still needed to dlopen.

No, we only need `dladdr`. The .hsaco is loaded through HIP Runtime `hipModuleLoadDataEx`.

We may need dlopen in the future (the planned runtime loader like glad) but the preferred solution is still to call Win32 API directly instead of relying on dlfcn-win32, which for some reason does not care about Unicode support.

jammm:
`dladdr` is part of dlcfn, no? So we still need dlfcn-win32 in that case.

EDIT: Ah no, this PR makes that redundant. So I'll get rid of dlfcn-win32.

include/aotriton/_internal/fd.h
```diff
@@ -0,0 +1,39 @@
...
+  return fd != INVALID_HANDLE_VALUE;
+}
+#else
+using fd_t = int;
+
+inline constexpr fd_t invalid_fd() {
+  return -1;
+}
+
+inline constexpr bool fd_is_valid(fd_t fd) {
```
xinyazhang: `constexpr` for `fd_is_valid`?
Well technically it can be but I can hardly image how it can be optimized over inline.

v3src/packed_kernel_unix.h
```diff
@@ -4,19 +4,24 @@
 #ifndef AOTRITON_V2_API_PACKED_KERNEL_UNIX_H
 #define AOTRITON_V2_API_PACKED_KERNEL_UNIX_H
 
+#include <aotriton/_internal/fd.h>
```
xinyazhang: Actually we don't need `packed_kernel_unix.h` or `packad_kernel_win32.h`
We can merge them into `fd.h`:
```cpp
#ifdef WIN32  // Find the right macro
#include "fd_win32.inl"
#elif POSIX
#include "fd_unix.inl"
#endif
```
Note the `.inl` suffix, which is recognized as cpp file (by vim at least)

v3src/packed_kernel_win32.h
```diff
@@ -4,53 +4,39 @@
...
-    // Regular paths - add \\?\ prefix
-    return LR"(\\?\)" + path;
-}
-
-// Convert HANDLE to int for our interface
-// We use intptr_t to ensure the HANDLE fits
-static int handle_to_fd(HANDLE handle) {
-    if (handle == INVALID_HANDLE_VALUE) {
-        return INVALID_FD;
+static std::wstring add_long_path_prefix(std::wstring path) {
```
xinyazhang: Optionally you can move complicated code implementation to fd_win32.cc/fd_unix.cc respectively. By decoupling fd related operations from packed_kernel and promoting them into designated module fd.h it may be worthy to get rid of the current head-only practice for potential new use cases.

xinyazhang (changes requested):
Still need some change. packed_kernel_<platform>.* can go into fd.h/fd_<platform>.inl/fd_<platform>.cc

jammm:
Will push your review fixes soon.

xinyazhang (changes requested):
No updates since my previous review

include/aotriton/_internal/fd.h
```diff
@@ -0,0 +1,53 @@
+// Copyright © 2026 Advanced Micro Devices, Inc.
+// SPDX-License-Identifier: MIT
+
+#ifndef AOTRITON_V3_FD_H
+#define AOTRITON_V3_FD_H
+
+#include <aotriton/config.h>
+#include <cstddef>
+#include <sys/types.h>
```
xinyazhang, line 10: `#include <stdio.h>` here

v3src/lszip.cc
```diff
@@ -42,7 +37,7 @@ static inline uint32_t le32(const uint8_t* p) {
 namespace AOTRITON_NS {
 
 bool
-lszip(int fd,
+lszip(fd_t fd,
       std::function<void(std::string_view, uint64_t, uint64_t)> visitor) {
   off_t file_size = fd_seek(fd, 0, SEEK_END);
```
xinyazhang, line 42: Compiler error: `‘SEEK_END’ was not declared in this scope`, need to include stdio.h at fd.h

xinyazhang (changes requested):
Missing `stdio.h`, otherwise good (tested locally).

include/aotriton/_internal/fd.h
jammm: added

jammm:
@xinyazhang added `stdio.h` include in fd.h.

xinyazhang (approved):

PR #186, opened 2026-06-16
AOTriton Template Instantiation

xinyazhang:
# Overview

This PR ship the **AOTriton Template Instantiation (ATI)** system: a
declarative Python DSL for describing Triton kernels and operators, plus a
two-pass compiler (Parser → shells, Linker → IR) that generates C++ shim
headers, operator dispatchers, and autotune LUTs without importing or executing
Triton.

The code generator moves from `v3python/` (largely deleted) to the new
`python/` tree, installed as the `aotriton` package. Flash kernel descriptions
move from `v3python/rules/flash/` to `modules/flash/aot/`; Triton kernel
sources move from `tritonsrc/` to `modules/flash/kernel/` with symlinks left in
`tritonsrc/` for backward compatibility.

Full user and developer documentation is in
`docs/AOTriton Template Instantiation.md` — read that first; this description
covers only the structural highlights.

## Major Changes

* [ati] Introduce the `python/template_instantiation/` package
  + Install: `pip install .` at root of source tree
  + Use: `import aotriton.template_instantiation as ati`

* [modular] Family kernel descriptions are now self-contained under
  `modules/<family>/`: ATI descriptions in `aot/`, Triton kernel sources in
  `kernel/`. This layout makes each family independently relocatable and
  eliminates the coupling between the code generator and, well, everything else.

* [codegen] Two-pass linker replaces the single-pass generator.
  + Pass 1 COMPILE (`python/codegen/parser.py`): loads each `modules/<family>/aot`
    package by explicit file path, reads passive decorator specs, emits
    lightweight shells with unresolved "relocations" (cite gaps, backend refs).
    - parser.py is used to avoid confusion with `python/compile.py`
  + Pass 2 LINK (`python/codegen/linker.py`): resolves `@ati.cite` gaps,
    builds all IR objects, infers `SHARED_IFACE`, derives operator params
    structs and default kernels, and emits the "undefined symbol" diagnostic (A4).

* [build] CMake migrated to the `aotriton` package (`python/` → `aotriton`).
  + All `python -m v3python.<tool>` invocations replaced by `python -m aotriton.<tool>`:
    - `aotriton.generate` — C++ shim/dispatcher/autotune code generation
    - `aotriton.compile` — single-kernel HSACO compilation
    - `aotriton.write_image_signature` — write HSACO entry-name metadata
    - `aotriton.aks2` — pack AKS2 blobs into an uncompressed ZIP archive
    - `aotriton.flatzip` — flat ZIP packaging for kernel images
    - `aotriton.database_compose` — compose tuning SQLite database
    - `aotriton.database_decompose` — decompose tuning SQLite database
    - `aotriton.comment_only_asm` — generate comment-only ASM stubs
  + `python/` is installed into the build venv via `pyproject.toml` / `setup.py`
    at configure time (non-editable install; no `.pth` leak).
  + Tuning database default root updated from `v3python/database/` to
    `modules/database/`.
    - This is a temporary location, the target directory is `modules/<family>/database`.
  + `--root_dir` argument added to `generate` and `write_image_signature` for
    path-independent module loading.
    - This argument makes it possible to generate AOT code for out-of-tree modules.

* [flash] Flash descriptions rewritten in ATI stacked-@ form under `modules/flash/aot/`.
  + `v3python/rules/flash/` deleted; `modules/flash/aot/` is the canonical location.
  + Triton kernel sources moved from `tritonsrc/` to `modules/flash/kernel/`;
    `tritonsrc/*.py` kernel files are now symlinks pointing there.

## Minor Changes

* [ati] Dead code removed (vulture + coverage analysis).

* [build] `.gitignore` anchored `*build*/` rule to the repo root to avoid
  false matches in subdirectories.

* [doc] `docs/AOTriton Template Instantiation.md` added: user guide (authoring
  surface, decorator reference, cite protocol) and developer guide (pipeline
  architecture, IR types, pass structure).

# Known Issues

* [ati] Tests live on a separate branch (`xinyazhang/ati-tests`) and are not
  included in this PR; the golden snapshot was re-baselined to full-ATI output
  but end-to-end CI for the new pipeline is pending.
* [db] Tuning databases under `modules/database/` are carried over from the
  previous location; no re-tuning was performed for this PR.
* [scope] This PR is the first stage of an ongoing modularization effort. The
  target end-state has exactly three top-level source directories that matter:
  `python/` (the `aotriton` package), `modules/` (per-family self-contained
  units), and `v3src/` (C++ code shared across all kernel families). Each
  `modules/<family>/` will grow subdirectories for every concern currently
  scattered across the tree:

  | Subdirectory  | Currently at                                       |
  |---------------|----------------------------------------------------|
  | `aot/`        | ✅ done (this PR)                                  |
  | `kernel/`     | ✅ done (this PR)                                  |
  | `tune/`       | `v3python/tune/<family>/`                          |
  | `test/`       | `test/`                                            |
  | `database/`   | `modules/database/amd/<arch>/<family>/`            |
  | `csrc/`       | `v3src/flash/` (family-specific C++)               |
  | `binding/`    | flash-specific code in `bindings/`                 |

  These migrations are deferred to subsequent PRs to keep this changeset
  reviewable.


modules/flash/aot/aiter_bwd.py
```diff
@@ -0,0 +1,49 @@
...
+It SUPPLIES the DQ_ACC operand (the workspace accumulator only this backend uses):
+the operator's params-struct union picks it up so DQ_ACC lands in OpAttnBwdParams
+without any hand-injection. DQ_ACC is a LazyTensor (rank 4, strides stride_acc?).
+"""
+
+import aotriton.template_instantiation as ati
+from ._common import check_value
+
+
+def _bwd_disabled(functional):
```
xinyazhang: rename to `_aiter_bwd_disabled`

modules/flash/aot/aiter_fwd.py
```diff
@@ -0,0 +1,41 @@
...
+`aiter::mha_fwd_args` cookie and packaging pre-built fmha_v3_fwd .co files. Declared
+declaratively via the @ati.affine.* surface; the operator (op_attn_fwd) it serves is
+referenced by name and resolved to SHARED_IFACE by infer_shared_iface.
+"""
+
+import aotriton.template_instantiation as ati
+from ._common import check_value
+
+
+def _fwd_disabled(functional):
```
xinyazhang: rename to `_aiter_fwd_disabled`

modules/flash/aot/attn_fwd.py
```diff
@@ -0,0 +1,133 @@
...
+import os
+from dataclasses import dataclass
+
+import numpy as np
+
+import aotriton.template_instantiation as ati
+from ._common import flash_disabled
+
+
+def _block_dmodel_values():
```
xinyazhang: move to _common.py and share

modules/flash/aot/attn_fwd.py
```diff
@@ -0,0 +1,133 @@
...
+    env = os.getenv('AOTRITON_FLASH_BLOCK_DMODEL',
+                    default='16, 32, 48, 64, 80, 96, 128, 160, 192, 224, 256, 512')
+    return [int(d) for d in env.split(',')]
+
+
+MAIN_DTYPES = ['*fp16:16', '*bf16:16', '*fp32:16']
```
xinyazhang: move to _common.py and share

modules/flash/aot/bwd_kernel_dk_dv.py
```diff
@@ -0,0 +1,92 @@
...
+import os
+from dataclasses import dataclass
+
+import numpy as np
+
+import aotriton.template_instantiation as ati
+from ._common import flash_disabled
+
+
+def _block_dmodel_values():
```
xinyazhang: move to _common.py and share

modules/flash/aot/bwd_kernel_dk_dv.py
```diff
@@ -0,0 +1,92 @@
...
+    env = os.getenv('AOTRITON_FLASH_BLOCK_DMODEL',
+                    default='16, 32, 48, 64, 80, 96, 128, 160, 192, 224, 256, 512')
+    return [int(d) for d in env.split(',')]
+
+
+MAIN_DTYPES = ['*fp16:16', '*bf16:16', '*fp32:16']
```
xinyazhang: move to _common.py and share

modules/flash/aot/bwd_kernel_dq.py
```diff
@@ -0,0 +1,76 @@
...
+import os
+from dataclasses import dataclass
+
+import numpy as np
+
+import aotriton.template_instantiation as ati
+from ._common import flash_disabled
+
+
+def _block_dmodel_values():
```
xinyazhang: move to _common.py and share

modules/flash/aot/bwd_kernel_dq.py
```diff
@@ -0,0 +1,76 @@
...
+    env = os.getenv('AOTRITON_FLASH_BLOCK_DMODEL',
+                    default='16, 32, 48, 64, 80, 96, 128, 160, 192, 224, 256, 512')
+    return [int(d) for d in env.split(',')]
+
+
+MAIN_DTYPES = ['*fp16:16', '*bf16:16', '*fp32:16']
```
xinyazhang: move to _common.py and share

modules/flash/aot/bwd_kernel_fuse.py
```diff
@@ -0,0 +1,61 @@
...
+
+import os
+from dataclasses import dataclass
+
+import numpy as np
+
+import aotriton.template_instantiation as ati
+
+
+def _block_dmodel_values_capped():
```
xinyazhang, line 23: move to _common.py and share

modules/flash/aot/bwd_preprocess.py
```diff
@@ -0,0 +1,49 @@
...
+
+import os
+from dataclasses import dataclass
+
+import numpy as np
+
+import aotriton.template_instantiation as ati
+
+
+def _block_dmodel_values():
```
xinyazhang: move to _common.py and share

modules/flash/aot/bwd_preprocess.py
```diff
@@ -0,0 +1,49 @@
...
+    env = os.getenv('AOTRITON_FLASH_BLOCK_DMODEL',
+                    default='16, 32, 48, 64, 80, 96, 128, 160, 192, 224, 256, 512')
+    return [int(d) for d in env.split(',')]
+
+
+MAIN_DTYPES = ['*fp16:16', '*bf16:16', '*fp32:16']
```
xinyazhang: move to _common.py and share

modules/flash/aot/bwd_preprocess_varlen.py
```diff
@@ -0,0 +1,45 @@
...
+
+import os
+from dataclasses import dataclass
+
+import numpy as np
+
+import aotriton.template_instantiation as ati
+
+
+def _block_dmodel_values():
```
xinyazhang: move to _common.py and share

modules/flash/aot/bwd_preprocess_varlen.py
```diff
@@ -0,0 +1,45 @@
...
+    env = os.getenv('AOTRITON_FLASH_BLOCK_DMODEL',
+                    default='16, 32, 48, 64, 80, 96, 128, 160, 192, 224, 256, 512')
+    return [int(d) for d in env.split(',')]
+
+
+MAIN_DTYPES = ['*fp16:16', '*bf16:16', '*fp32:16']
```
xinyazhang: move to _common.py and share

modules/flash/aot/flash_entry.py
```diff
@@ -0,0 +1,38 @@
...
+The tuning-side parse/dacite helpers stay in v3python.tune; keeping this copy free
+of v3python.tune is what severs the last v3python edge from the codegen closure.
+
+`as_text()` output MUST stay byte-identical to the original (it is the diagnostic
+line the tuner re-parses).
+"""
+
+from dataclasses import dataclass, asdict
+
+
```
xinyazhang, line 20: Add "TODO: Merge with modules/flash/tune in ATI Phase 2: Modularization"

python/template_instantiation/ir/choice.py
```diff
@@ -0,0 +1,129 @@
...
+  still depend on, mid-migration, while the golden gate requires the old path to
+  stay byte-for-byte identical. So Choice is a deliberately TEMPORARY adapter.
+  Plan: keep Choice until the migration is complete; then merge its features into
+  TypedChoice and delete this wrapper. See executive plan Step 6.3.
+"""
+
+from . import typed_choice as TC
+
+
+class Choice:
```
xinyazhang: Now we completed the migration of v3python -> python as aotriton package. We should remove this wrapper layer

python/template_instantiation/ir/functional.py
```diff
@@ -0,0 +1,296 @@
...
+    @property
+    def noptimized_for(self):
+        return len(self._optimized_for)
+
+    @property
+    def family(self):
+        return self.meta_object.FAMILY
+
+    @property
+    def FAMILY(self):
```
xinyazhang: unify the usage across the code base. All lowercase. Upper case is reserved for anticipated constants which makes sense when subclassing v3python's KernelDescriptor/Interface, but makes no sense for Functional objects.

python/template_instantiation/ir/interface.py
```diff
@@ -0,0 +1,116 @@
...
+        its DEFAULT backend's axes while keeping meta_object = the operator."""
+        return self._built.axes, self._built.overrides
+
+    def gen_functionals(self, target_arch):
+        """Yield every Functional of this Interface (the classical enumeration:
+        product over the multi-choice axes (godel), fan each variable's choice onto
+        its arguments, apply overrides in declared order). `target_arch` is an
+        ordered {arch -> gpus}; arch_number is its enumeration index."""
+        import itertools
+        from .axis import assign_godel, godel_of
```
xinyazhang: move to top of the module if lazy loading is not desired.

python/template_instantiation/ir/kdesc.py
```diff
@@ -0,0 +1,612 @@
...
+                    self._autotune_keys_validated.append((key, _binning_class(sel)))
+
+    # --- perf params + tuning metadata (legacy translate_* contract) ---
+
+    @property
+    def AUTOTUNE_KEYS_VALIDATED(self):
+        return self._autotune_keys_validated
+
+    @property
+    def PROGRAMMATIC_PERFS(self) -> dict:
```
xinyazhang: Should have retired.

python/template_instantiation/ir/kdesc.py
```diff
@@ -0,0 +1,612 @@
...
+        if ts is not None:
+            argset = set(built.arguments)
+            for key, sel in ts.binning.items():
+                if key in argset:
+                    self._autotune_keys_validated.append((key, _binning_class(sel)))
+
+    # --- perf params + tuning metadata (legacy translate_* contract) ---
+
+    @property
+    def AUTOTUNE_KEYS_VALIDATED(self):
```
xinyazhang: Should have retired.

python/template_instantiation/ir/kdesc.py
```diff
@@ -0,0 +1,612 @@
...
+        from aotriton.codegen.parser import load_family_aot
+        FlashKernel = load_family_aot('flash')._common.FlashKernel
+        return FlashKernel.sancheck_lut_tensor(self, f, lut_tensor)
+
+    def _gen_missing_entries(self, *args, **kwargs):
+        from aotriton.codegen.parser import load_family_aot
+        FlashKernel = load_family_aot('flash')._common.FlashKernel
+        return FlashKernel._gen_missing_entries(self, *args, **kwargs)
+
+    def update_programmatic_perfs(self, kw, f):
```
xinyazhang: Should have retired.

python/template_instantiation/ir/ops/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright © 2026 Advanced Micro Devices, Inc.
+# SPDX-License-Identifier: MIT
+
+from .union import union_params
+from .infer import infer_shared_iface
+from .cite import resolve_cites
+
+__all__ = ['union_params', 'infer_shared_iface', 'resolve_cites']
```
xinyazhang: I didn't see the necessity to put this package directly under template_instantiation. If not strictly needed put this package under ir/

python/template_instantiation/decorators.py
```diff
@@ -0,0 +1,662 @@
+# Copyright © 2026 Advanced Micro Devices, Inc.
```
xinyazhang: This file is too long. Make it a package and with one file one decorator

test/ati/golden/files/flash/iface.op_attn_fwd.h
```diff
@@ -0,0 +1,129 @@
+// Copyright © 2025 Advanced Micro Devices, Inc.
```
xinyazhang: Do not include test/ati into the final PR. Create a different worktree+branch to keep them.

python/template_instantiation/metro/build.py
```diff
@@ -0,0 +1,138 @@
...
+
+class MetroKernel(Interface):
+    """A metro launcher built from a transpiled @ati.metro_kernel plan. Has no
+    functional space of its own — it sequences sub-kernel contexts."""
+
+    CODEGEN_MODULE = 'op'
+    TUNE_NAME = None
+    ENUM_PREFIX = 'kMetro_'
+
+    def __init__(self, name, kernels, *, family='flash'):
```
xinyazhang: No implicit family='flash' throughout the codebase.

python/codegen/parser.py
```diff
@@ -0,0 +1,240 @@
...
+its `kernel/` sources keep importing each other by bare name.
+"""
+
+import os
+import sys
+import importlib.util
+from pathlib import Path
+
+
+def _resolve_modules_dir() -> Path:
```
xinyazhang: This doesn't work in general case. where modules can be placed at any directory. parser should use the directory passed in by `aotriton.generate --root_dir`, not guessing the modules from __file__

jeffdaily (changes requested):
## Review — AOTriton Template Instantiation

### Summary
Ships the ATI system: a declarative Python DSL plus a two-pass (parser->linker) Triton-free code generator that emits C++ shims, operator dispatchers, and autotune LUTs, and relocates the generator (`v3python/`->`python/`) and kernel sources (`tritonsrc/`->`modules/flash/kernel/`, with BC symlinks). The architecture is sound and the design doc is excellent, but the PR ships an ~8K-LOC generator rewrite **with no codegen regression test** (the golden-snapshot harness was deliberately removed from the branch) and leaves several correctness gaps. Requesting changes.

### Testing (blocking)
- **No codegen regression test ships.** `test/ati` is absent (commit "remove test/ati from the PR branch (kept on xinyazhang/ati)") and nothing under `test/` exercises the generator. The entire design's correctness criterion is *golden-identical output*; the one harness that verifies it was stripped. A reshuffle this large has zero protection against output drift. Re-add a snapshot/golden test, or justify its absence.
- **`test/test_gpu_targets.py` is broken.** All 5 parametrized cases do `python -m v3python.gpu_targets` (lines 34/51/60/71/80), but `gpu_targets.py` moved to `python/` (`aotriton.gpu_targets`); `v3python/gpu_targets.py` no longer exists, and CMake now invokes `-m aotriton.gpu_targets` (`v3src/CMakeLists.txt:17`). Every case will fail with `ModuleNotFoundError`. Re-point the `@pytest.mark.parametrize("module", [...])` to `aotriton.gpu_targets`.

### API Design / Correctness — Needs Discussion
- **`bwd_kernel_fuse` constexpr vs runtime divergence.** `num_head_q/k`, `num_seqlens`, `max_seqlen_q/k` are `tl.constexpr` in `modules/flash/kernel/bwd_kernel_fuse.py:67-73`, but the fuse aot (`modules/flash/aot/bwd_kernel_fuse.py:43-45`) inherits them via `@ati.cite` from `bwd_kernel_dk_dv`, where they are runtime `@ati.scalar(..., 'i32')`. Root cause: `introspect.py:56` hardcodes `is_constexpr=False` for all sourced-kernel params, so the ATI description -- not the kernel source -- is authoritative and **nothing cross-checks the divergence**. If the fuse backend is compiled/launched with these as runtime args, the launch signature won't match the constexpr-specialized binary. Please confirm the fuse kernel specializes correctly, or redeclare these five locally as constexpr (`options=`/`scalar_var`) in the fuse aot.

### Code Quality / Correctness
- **`python/codegen/autotune.py:73`** -- `assert all([isinstance(k, KernelSignature)] for k in self._sigs)` is a no-op: each element is a one-element (always-truthy) list, so the type check never runs. Fix to `all(isinstance(k, KernelSignature) for k in self._sigs)`.
- **`python/codegen/autotune.py:169-172`** and **`ir/typed_choice.py:358-366` (`GuessInt`)** -- both size an integer C type from `max(value)` only, ignoring negative minima, and `autotune` uses strict `<` with no out-of-range guard (silently emits `int64_t` if nothing fits). A LUT/choice list like `[-40000, 0]` selects `int8_t` and truncates. Size on `max(abs(min), abs(max))` and raise if no type fits. Latent today (flash values are small/non-negative).
- **`python/utils/lazy_file.py:35-43`** -- `LazyFile.__exit__` ignores `exc_type` and flushes its in-memory buffer unconditionally. If `write_shim_source` raises mid-generation (`codegen/interface.py:108`), the already-written `.h` plus a partial `.cc` are left on disk -- a corrupt, non-deterministic pair that breaks the golden-output invariant. Skip the write when `exc_type is not None`.
- **`ir/ops/union.py:62`** (`build_merged_struct_cfields`) -- a backend `union_order` anchor whose cfield is supplied by another sub-kernel raises a bare `KeyError` (confirmed via repro) instead of a `DescriptionError` naming the operand. Crash-quality regression on a real authoring mistake; guard the lookup.
- **`codegen/common.py:45`** -- `max([len(cf.ctype) for cf in cfields])` raises `ValueError` on an empty derived struct rather than emitting an empty struct.
- **`registry.py`** -- module-global `_KERNELS`/`_OPS` with no reset between runs: `register_kernel` `setdefault(...)[NAME] = kdesc` silently stomps a duplicate instead of erroring, and `clear()` has zero non-test callers. A long-lived process (test session, double invocation) gets order-dependent cite resolution. Have the generation driver `clear()` per family and make duplicate registration assert.
- **`ir/ops/cite.py` (path 1) / `registry.py:38` `register_op`** -- `register_op` is never called, so the `get_op` ops-registry branch in `_resolve_one_cite` is unreachable dead code (real cites resolve via `metro_lookup` and the flat registry). Remove the dead branch or wire `register_op`.

### Build / Packaging
- **v3python retirement is incomplete/contradictory.** `git ls-files v3python` shows 65 still-tracked files (the whole `tune/` subtree), yet the commit message claims retirement and `.gitignore:22-25` ignores a non-existent `v3python.retiring/`. No build/runtime code imports `v3python` (clean), but the state contradicts the narrative. Decide: keep with a one-line note that `v3python.tune` is the not-yet-migrated tuning path, or remove it.
- **Installed wheel is incomplete.** `setup.py:48-53` ships only `aotriton.codegen` template `.cc/.h`; it does not ship `modules/` (kernel sources, aot descriptions) or `modules/database/**`. `python -m aotriton.generate` from an installed wheel fails at link time (`codegen/root.py:77,303` read from `root_dir/modules`). Not a build break (CMake passes `--root_dir <source>`), but it contradicts pyproject's "installable generator" intent. Either package `modules/**`+database or document the wheel as build-venv-only.
- **`tools/preview.py:18`** uses absolute `from aotriton.codegen.common import ...` while siblings use relative imports; hard-codes the top-level package name. Make it relative.

### Robustness / Cleanup
- **`introspect.py:56`** also drops default arg values (no `a.defaults` capture) with no signal -- a future sourced kernel with a defaulted constexpr silently becomes a required runtime param. Capture at least a `has_default` flag.
- **`introspect.py:62-67`** -- the AST path appends `kwonlyargs` last (ABI order) but the `inspect.signature` fallback emits declaration order; the two disagree on ARGUMENTS order for a `*`-separated signature.
- **`_common.py:64-76` vs `21-37`** -- `flash_disabled` and `FlashKernel.is_functional_disabled` are two copies of "is this functional disabled" that already disagree on the gfx950 bad-hdim exclusion; the LUT sancheck can expect entries the build excludes. Consolidate.
- **NUM_XCDS=8 arch list** is hardcoded to `('gfx942','gfx950')` and duplicated per kernel (config + `@ati.derives`) in `attn_fwd.py`, `bwd_kernel_dk_dv.py`, `bwd_kernel_dq.py`; a future multi-XCD arch silently gets `NUM_XCDS=1`. Centralize in `_common.py`.
- **`bwd_kernel_dk_dv.py:8`** docstring names `../kernel/bwd_split_kernel.py` but the actual `@ati.source` (line 80) is `bwd_kernel_dk_dv.py`; `bwd_split_kernel.py` has no top-level FunctionDef, so a "fix to match the comment" would break. Fix the docstring.
- **`ir/ops/cite.py:268` / `_string_dtype_names`** returns a `set`, leaking iteration order into `spec.dtype_vars` append order (currently masked by the anchor re-sort). Return sorted/spec-order.
- **No empty-choice guard** -- `options=[]`/`dtype=[]` yields a radix-0 axis and later `choices[0]` -> `IndexError`. Validate `len(choices) >= 1` at spec/axis construction.
- **Load-bearing TODO/FIXMEs left in generated control flow:** `codegen/operator.py:127` ("lookup all then launch" -- metro launcher launches before confirming all sub-kernels resolve), `codegen/root.py:137` (Windows `validator_function` KeyError), plus a dead commented `write_cluster`. Resolve or track explicitly.
- **`linker.py:60`** -- a 3-segment cite naming a kernel absent from `compiled.kernels` produces no dependency edge and only a late/confusing error; error early with "cite target not found."
- **`decorators/affine.py` `SuppliesSpec`** -- multiple `@ati.affine.supplies` with differing `after=`/`before=` silently last-wins (`specs/affine.py:74-77`); reject contradictory anchors.

### Recommendation
**Request Changes.** Blocking: missing codegen regression coverage (the design's correctness is defined as golden-identical output, yet the verifying harness was removed) and the broken `test_gpu_targets.py`. The `bwd_kernel_fuse` constexpr/runtime divergence needs author confirmation before merge -- it is a plausible functional bug that no validation in the pipeline would catch. The remaining code-quality items (no-op assertion, partial-write on exception, integer-width sizing, dead registry paths, v3python/packaging contradictions) should be fixed in this PR.

Note: two findings from an initial pass were dropped after verification -- the affine `@ati.start` "mis-dispatch" (the `aiter_asm` marker *is* innermost, so `specs[-1]` is correct) and `register_op` "breaking operator cites" (downgraded to dead code; real cites resolve via `metro_lookup`/flat registry).

_Reviewed with the assistance of Claude Code._

xinyazhang:
@jeffdaily fixed all

# Replies to Review 4535159912

## Fixed

**B2 — test_gpu_targets.py wrong module path**
Fixed in 4d2741cc.

**C1 — autotune.py:73 no-op assert**
Fixed in 4d2741cc (removed spurious `[...]` wrapper).

**C2 — bwd_kernel_dk_dv.py docstring wrong filename**
Fixed in 4d2741cc.

**D1 — GuessInt negative handling / LUT dtype selection**
`GuessInt`: abs(v) is intentional conservative margin; added a comment.
`autotune.py:169-172`: fixed in e0b669a0 to check both min and max.

**D2 — lazy_file.py flushes on exception**
Fixed in 4d2741cc (`if exc_type is not None: return` in `__exit__`).

**D3 — common.py max() on empty cfields**
Fixed in 4d2741cc (early return). Empty cfields is valid for untunable
kernels and metro launchers.

**D5 — union.py bare KeyError**
Fixed in 2293adef with operand name, full merged order, defined
operands, and sub-kernel types.

**E9 — register_op / get_op dead code**
Used only by tests. Moved to `xinyazhang/ati-tests` in 543094b3;
deleted from the aotriton package entirely.

**E10 — options=[] silent radix-0**
Fixed in 4d2741cc (assertion in `Axis.__init__`).

**E12 — linker.py late cite error**

**E13 — SuppliesSpec conflicting anchors**
Fixed in aec79b6b44d064834c2c61ec06de0e69be67cf3a

## Intentional Design

**B1 — No regression test in this PR**
Tests are on `xinyazhang/ati-tests`; snapshots would bloat this PR.
39 tests pass; pointer added to the PR description.

**D4 — registry.py no reset**
Moved to ati-test

**D6 — `_string_dtype_names` returns set**
Not a problem. The caller resolves each name from a dict keyed by
name, so append order in spec.dtype_vars is irrelevant.

**E1 — bwd_kernel_fuse constexpr/runtime**
ATI is triton-free: `is_constexpr` comes from `options=`, not the Triton
source. Runtime decl is correct; Triton JIT-specializes on concrete values.

**E2 — introspect.py drops stub defaults**
Stub defaults are irrelevant; all type info comes from `@ati.*` decorators.

**E3 — kwonlyargs order disagreement**
Production uses KernelStub AST only (Triton kernels have no kwonlyargs).

**E4 — 65 v3python files remain**
All 65 are `v3python/tune/`, kept intentionally as Stage 1 (see PR desc).

**E5 — Installed wheel missing modules/**
Intentional, modules is not part of the package.

**E6 — preview.py absolute import**
Intentional: run as `python -m aotriton...`; relative import would break.

**E7 — flash_disabled vs is_functional_disabled**
TODO added before `FlashKernel` in 8aa4c94c; will unify when tune/ is
modularized.

**E8 — NUM_XCDS arch list duplicated**
Deliberate: each description is self-contained. Clean follow-up, not here.

**E11 — Load-bearing FIXMEs**
Neither affects Linux correctness; one is a perf note, one a Windows gap.

PR #188, opened 2026-06-24
fix(tuning): fall back to hwmon when GPU temp unavailable through amdsmi

amd-callumm:
Co-authored by: Claude. Reviewed and signed off by Callum Mitchell.

## Motivation

amdsmi_get_temp_metric returns a exception on gfx1151 where GPU temperature sensing is not yet supported (I believe this is a limitation of amdsmi/not yet implemented):
```
File "/opt/venv/lib/python3.12/site-packages/amdsmi/amdsmi_interface.py", line 764, in _check_res
    raise AmdSmiLibraryException(ret_code)
amdsmi.amdsmi_exception.AmdSmiLibraryException: Error code:
        2 | AMDSMI_STATUS_NOT_SUPPORTED - Feature not supported
```

This was causing any tuning jobs scheduled for the platform to fail. This PR adds temperature sensing through hwmon as a backup for such cases. This will fix Linux, but not Windows, which will proceed without

## Technical Details

Updates v3python/tune/gpu_utils.py to fall back to hwmon when amdsmi fails to obtain temperature information. This allows for (Linux) tuning machines to monitor GPU temperature. Junction temperature will be prioritized if available (same as amdsmi function), otherwise edge temperature will be used.

Since hwmon is Linux-native, it will also fail on Windows, and the tuning will continue without temperature monitoring (with a warning logged once). Warnings will also be logged if the hwmon fallback fails for any other reason.

## Test Plan

- In a local python environment, called _get_temperature, _get_temperature_amdsmi, and _get_temperature_hwmon on a Linux gfx1151 machine to verify expected behavior and workings
- Worked through tuning DB setup steps, scheduled and ran jobs on a gfx1151 Linux worker

## Test Result

Local python calls show expected behavior.

On Linux, gfx1151 tuning jobs that failed with before these changes now succeed, and their logs no longer show this error.

**Note:** Windows and non-gfx1151 behavior are currently **untested**.

### DIFF
--- a/v3python/tune/gpu_utils.py
+++ b/v3python/tune/gpu_utils.py
@@ -1,6 +1,7 @@
 # Copyright © 2025 Advanced Micro Devices, Inc.
 # SPDX-License-Identifier: MIT
 
+import logging
 import sys
 import os
 import time
@@ -32,6 +33,8 @@ from .defaults import (
     default_device_id,
     default_device_string,
 )
+
+logger = logging.getLogger(__name__)
 
 def elike(t: torch.Tensor | None) -> torch.Tensor | None:
     return torch.empty_like(t) if t is not None else None
@@ -85,6 +88,8 @@ def detach_member_tensors(data_object) -> dict:
 
 _amdsmi_initialized = False
 _hip_to_amdsmi = {}
+_hip_to_drm_card = {}
+_hwmon_warning_logged = False
 _total_memory_gb = None
 
 def get_total_memory_from_amdsmi():
@@ -110,7 +115,7 @@ def get_total_memory_from_amdsmi():
 
 def _init_amdsmi():
     """Initialize AMD-SMI and build HIP to AMD-SMI device mapping."""
-    global _amdsmi_initialized, _hip_to_amdsmi
+    global _amdsmi_initialized, _hip_to_amdsmi, _hip_to_drm_card
     if _amdsmi_initialized:
         return True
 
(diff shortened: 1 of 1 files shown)


xinyazhang (changes requested):
This is unnecessary, `amdsmi` package is already guaranteed by https://github.com/ROCm/aotriton/blob/main/.tune/image.scripts/02-install_amdsmi.sh
In addition, we have the control over the tuning environment through containers and there is no need to support all environments.

However you may want to make the failure loud. The malfunctioning of temperature reporting can cause weird bug.

amd-callumm:
> This is unnecessary, amdsmi package is already guaranteed by https://github.com/ROCm/aotriton/blob/main/.tune/image.scripts/02-install_amdsmi.sh
In addition, we have the control over the tuning environment through containers and there is no need to support all environments.

The issue isn't with the amdsmi package being missing; it's that even with amdsmi present, installed and working within the tuning environment/container, the amdsmi package doesn't yet support temperature reading on certain GPU targets. On gfx1151 it returns "N/A" for all sensors - this is a [known and open issue](https://github.com/ROCm/ROCm/issues/6035) with amdsmi (I'm not sure if/which other targets are affected).

Ideally, this issue gets fixed soon. Until then, we need some way to monitor the temperature of the GPU target being tuned for, and hwmon provides this.

xinyazhang:
Ultimately this is not an amd-smi problem, it's an `amdgpu` problem. On gfx1151 system you can see
```
$ find /sys/class/drm/card*/device/hwmon/ -name 'temp*'
/sys/class/drm/card0/device/hwmon/hwmon3/temp1_label
/sys/class/drm/card0/device/hwmon/hwmon3/temp1_input`
```
And `head /sys/class/drm/card*/device/hwmon/hwmon*/temp1_*` will give things like
```
==> /sys/class/drm/card0/device/hwmon/hwmon3/temp1_input <==
30000

==> /sys/class/drm/card0/device/hwmon/hwmon3/temp1_label <==
edge
```

This means hwmon approach does not give any extra information than amd-smi, and it is also non-trivial to support hwmon.

Thus the easiest fix is to pull `amdsmi.AmdSmiTemperatureType.EDGE` as a fallback of `amdsmi.AmdSmiTemperatureType.JUNCTION`, rather than going through sysfs which provides zero additional information.

PR #203, opened 2026-07-22
Implement partial database support through GPU priority lists

amd-callumm:
## Motivation

Implement the TODO in sqlite.py (lines 59-61) to enable partial tuning databases where each GPU target constructs a tuning LUT based on a prioritized list of GPU databases to draw from.

This allows targets with incomplete tuning databases to use their partial results, while falling back to a complete, reliable database for shapes not yet tuned, enabling incremental deployment of tuning results.

e.g. - gfx1151 has [gfx1151_mod0, gfx1100_mod0] as its list, so for each data configuration, it will check the gfx1151 tuning database for a kernel mapping and use it. If there is no entry for the config, it will fall back to the next priority (gfx1100 database). Lists can contain any number of entries.

Different variants of the same target can also be requested, e.g. (gfx942mod2, gfx942mod1, gfx942mod0) for gfx942.

## Technical Details

Changes:
- Define AOTRITON_TUNING_DATABASE_REUSE as dict[str, list[str]] mapping target GPU → priority chain (e.g., 'gfx1151_mod0': ['gfx1151_mod0', 'gfx1100_mod0']) in gpu_targets.py
- Change database_gpus property to return dict[str, list[str]] with per-target priority chains (functional.py)
- Implement SQL-based priority selection in sqlite.py:
  - Build UNION ALL query with one branch per target GPU
  - Each branch uses ROW_NUMBER() OVER (PARTITION BY inputs$* ORDER BY CASE gpu ...) to select highest-priority row per config
    * Add target_gpu column to identify which target each row serves
    * Factor choice WHERE conditions into filtered_base CTE to evaluate once
    * Cache inputs$ column names per table (_input_config_cols_cache)
  - Update LUT generation to filter DataFrame by target_gpu column
    (kdesc.py, operator.py)

## Test Plan

- Local build of AOTriton with the changes in place (targeting gfx1151 + 1150)
- Manual verification that generated LUTs map to expected gfx1151/1150-tuned kernel parameters, where present
- Similar for gfx942 with different _mod* suffix

## Test Result

gfx1150 + gfx1151 build:
- New LUTs generated with the expected contents/format
- LUT entries match expected kernel parameters (including gfx1100 fallback for missing shapes)

gfx942 multi-mod build:
- Build targets: gfx942_mod0;gfx942_mod1
- Temporary _mod1 database used containing a subset of _mod0 input shapes, with different kernel parameters
- Temporary priority list of `'gfx942_mod1': ['gfx942_mod0']`
- Correct LUT generated with working DB fallback for input shapes missing from _mod1

### DIFF
--- a/python/database/sqlite.py
+++ b/python/database/sqlite.py
@@ -43,6 +43,7 @@ class Factory(object):
         log(lambda : f'sqlite3.connect({path / self.SIGNATURE_FILE})')
         self._conn = sqlite3.connect(path / self.SIGNATURE_FILE)
         self._conn.set_trace_callback(log) # Debug
+        self._input_config_cols_cache = {} # Cache "inputs$" columns in database
         for schema, bn in self.SECONDARY_DATABASES.items():
             fn = path / bn
             if fn.is_file():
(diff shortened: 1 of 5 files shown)


python/database/sqlite.py
```diff
@@ -56,29 +56,51 @@ def create_view(self, functional):
...
         table_name = pfx + meta.FAMILY.upper() + '$' + meta.NAME
-        # TODO: Incremental changes:
-        # 1. load database_gpus first
-        # 2. then override entries with optimized_for gpus
-        def build_sql(choice_dict):
-            wheres = {
-                'gpu' : functional.database_gpus,
-            }
+
+        # Extract GPU lists from tuples: database_gpus now returns [(target_gpu, fallback_gpu), ...]
```
xinyazhang: This design can be made even more generic: we don't have to explicitly list a (target, fallback) tuple. Instead we this code can support a list of database gpus like (tier0gpu, tier1gpu, tier2gpu) etc., which allows more flexible pattern of composition.

A concrete example is we can request (gfx942mod2, gfx942mod1, gfx942mod0) to get a 3-tier composition.

Thus with this new design there is no need to rename the `AOTRITON_TUNING_DATABASE_REUSE` to `AOTRITON_TUNING_DATABASE_FALLBACK` because there is no single "fallback".

python/database/sqlite.py
```diff
@@ -56,29 +56,51 @@ def create_view(self, functional):
...
+        target_gpus = [target for target, _ in functional.database_gpus]
+        fallback_gpus = [fallback for _, fallback in functional.database_gpus]
+
+        def build_sql(choice_dict, gpu_list):
```
xinyazhang: It is possible to achieve the shadowing directly with SQL statement, example:
```SQL
WITH ranked AS (
    SELECT
        gpu, feat1, feat2, data1, data2,
        ROW_NUMBER() OVER (
            PARTITION BY feat1, feat2
            ORDER BY CASE gpu
                WHEN 'gfx942mod2' THEN 1
                WHEN 'gfx942mod1' THEN 2
                WHEN 'gfx942mod0'  THEN 3
                ELSE 99
            END
        ) AS rn
    FROM table
)
SELECT feat1, feat2, data1, data2
FROM ranked
WHERE rn = 1;
```
It is supported since sqlite 3.25.0 (release in 2018) and we can safely use it.

xinyazhang (changes requested):
Two major things:
1. Make the "fallback" generic, we should support requesting a precedence of GPUs
2. The "fallback" can be done in SQL directly.

python/database/sqlite.py
```diff
@@ -51,32 +52,106 @@ def __init__(self, path):
...
-        def build_sql(choice_dict):
-            wheres = {
-                'gpu' : functional.database_gpus,
-            }
+
+        # Get all target GPUs and their priority chains
+        target_priority_map = functional.database_gpus  # dict[str, list[str]]
+
+        def build_sql_with_window_functions(choice_dict):
+            """Build UNION ALL query with window functions for each target GPU."""
```
amd-callumm, line 83: This UNION ALL approach is to support priority lists with multiple mod* entries for the same gfx target, such as (gfx942_mod2, gfx942_mod1, gfx942_mod0) for gfx942. I assume from PR comments that this is desirable functionality, even though it currently appears to be unused.

If this is overkill (i.e. if we can assume each target's priority list has only one entry per GPU/one database per GPU), we can likely simplify the SQL changes here.

xinyazhang (changes requested):
This is a larger changeset than my anticipation. We probably don't need to enumerate the input column names to achieve this goal.

python/gpu_targets.py
```diff
@@ -22,18 +22,20 @@
...
-    'gfx1101_mod0' : 'gfx1100_mod0',
-    'gfx1102_mod0' : 'gfx1100_mod0',
-    'gfx1103_mod0' : 'gfx1100_mod0',
-    'gfx1200_mod0' : 'gfx1201_mod0',
-    'gfx1150_mod0' : 'gfx1100_mod0',
-    'gfx1151_mod0' : 'gfx1100_mod0',
-    'gfx1152_mod0' : 'gfx1100_mod0',
-    'gfx1153_mod0' : 'gfx1100_mod0',
-    'gfx1250_mod0' : 'gfx942_mod0',
+    'gfx1101_mod0': ['gfx1101_mod0', 'gfx1100_mod0'],
```
xinyazhang: This is verbose. The precedence can be easily inferred for any gpu:
* self (exact arch+mod)
* exact arch, mod0
* The one specified by `AOTRITON_TUNING_DATABASE_REUSE`

Adding identity as first element does not add extra information and only make the logic more error prone.

python/database/sqlite.py
```diff
@@ -52,33 +52,67 @@ def __init__(self, path):
...
-            log(lambda : f'Table {table_name} may not exist. select stmt: {stmt} params {params}')
-            return None, format_sql(stmt, params)
+            """Query each target GPU's priority chain, return combined results.
+
+            Returns (dataframe, sql) where dataframe has target_gpu column.
+            """
+            all_dfs = []
+            last_sql = None
+
+            for target_gpu, priority_chain in target_priority_map.items():
```
xinyazhang: As I replied before, this function should be done in SQL statement instead programmatically with python.

xinyazhang (changes requested):

amd-callumm:
The last push was a rebase.

PR #200, opened 2026-07-13
Add gfx1150 and gfx1151 flash tuning databases for hdim=64, dropout=0.0

amd-callumm:
## Motivation

Tune high-priority flash attention shapes on gfx1150 + gfx1151. Includes more data types and shapes than #194 in addition to the second target architecture.

Current coverage is limited to the most immediately valuable shapes for my team's use cases/priorities. The full database will come in stages since it will take a long time to run everything on our worker machines.

## Technical Details

Tuning database coverage:
  - Head dimension: 64
  - All supported data types and sequence lengths
  - dropout=0.0 only

Performance improvements (attn_fwd kernel vs. default):
  - gfx1150 average speedups: fp16=64.7%, bf16=66.2%, fp32=76.3%
  - gfx1151 average speedups: fp16=61.2%, bf16=61.0%, fp32=71.1%

## Testing

Complete: verified no tuning failures or missing entries based on the above scope
Pending: build and run tuning test libraries and run level 0 and 1 test suites

Issue #183, opened 2026-06-05
attn_fwd fp32 hdim80 causal produces NaN in LSE on gfx1150 (clean on gfx1151)


mgehre-amd:
## Summary

The AOTriton-shipped fp32 forward Flash-Attention kernel for `hdim=80, causal=True` on gfx1150 (Strix Point) produces non-deterministic `NaN` values in the LSE (`M`) output tensor at a rate of ~50% per call on the shape `(B=3, H=5, Sq=256, Sk=32, D=80)`. The `Out` tensor is never corrupted. The exact same Triton source compiled for gfx1151 (Strix Halo) is clean. NaN positions are always paired in adjacent Sq positions `(2N, 2N+1)`.

Root cause is localised to gfx1150 hsaco. Cross-loading proves the race travels with the binary:

- `gfx1150-built hsaco` races on both gfx1150 (53% NaN rate) and gfx1151 (74% NaN rate)
- `gfx1151-built hsaco` is clean on both chips (0/500 trials)

Recompiling the gfx1150 hsaco with `-mllvm -amdgpu-enable-vopd=0` seems to avoid the bug.

Tested against AOTriton tag `0.12b` (SHA `2690368`), bundled Triton submodule `db82b800c`, torch `2.13.0a0+rocm7.14.0a20260527`.

## Reproducer

Standalone single-file reproducer using only `pyaotriton + torch`. No AOTriton test harness, no Triton JIT.

Assuming a built+installed AOTriton 0.12b at `$AOTRITON_INSTALL/lib/`:

```bash
source <venv>/bin/activate
export AOTRITON_TEST_LIBDIR=$AOTRITON_INSTALL/lib
export PYTHONPATH=$AOTRITON_TEST_LIBDIR
python minimal_repro_race.py 500    # exits 1 on race; expected on gfx1150
```

`minimal_repro_race.py`:

```python
#!/usr/bin/env python
"""Standalone reproducer for the gfx1150 fp32 attn_fwd LSE race."""
import sys, torch
from pyaotriton import T0, T1, T2, T4, DType, Stream
from pyaotriton.v3.flash import attn_fwd, attn_fwd_params

CAUSAL_TYPE_WINDOWED = 3
WINDOW_TOP_LEFT_ALIGNED = -2147483647

BATCH, N_HEADS, SEQLEN_Q, SEQLEN_K, D_HEAD = 3, 5, 256, 32, 80

[...]

PR #205, opened 2026-07-27
Add expanded gfx1151 and gfx1150 tuning databases

amd-callumm:
## Motivation

Expands the partially-complete gfx1151  and gfx1150 tuning databases (hdim=64 only) from #200 to include all supported data shapes, types and configurations (note: tuning configurations exclude seqlen 4096 and 8192, as well as hdim > 256).

To take advantage of these databases for supported shapes, #203 (partial tuning database/fallback support) will also be needed.

## Testing

Correctness tests Level 0 + Level 1 with local gfx115x builds on capable hardware.
- 100% passed on gfx1151
- A few accuracy/NaN issues on gfx1150, likely related to #183 since the same hdim + fp32 combo is affected
- adiff included in commit, but it varies a bit from run to run.


modules/flash/tests/adiffs/gfx1150.txt
```diff
@@ -0,0 +1,57 @@
+modules/flash/tests/test_backward.py::test_regular_bwd[Split-False-l1-dtype2-0.0-CausalOn-16-256-hdim80-5-3] (call)	NAN
```
xinyazhang, line 1: Although supported, adiff file should only be used to record NaN. Any NaN is a compiler/db/kernel bug and should be fixed.

xinyazhang (commented):
This PR requires https://github.com/ROCm/aotriton/pull/203 to work.

PR #209, opened 2026-08-13
Harden AKS2 packed-kernel parser against malformed input

fjankovi:
## Summary

`PackedKernel::PackedKernel()` trusts every field of the AKS2 header and of each directory entry. The per-kernel directory walk is not bounded against the decompressed payload, so `number_of_kernels` / `filename_length` can run `parse_ptr` off the end of `decompressed_content_`, and `filter()` can return `kernel_start_ + offset` pointing outside the buffer — a pointer that goes on to `hipModuleLoadDataEx()` in `v3src/on_device_kernel.cc:84`, which takes no length and parses the ELF from wherever it lands.

The existing trailing check

```cpp
if (kernel_start_ - decompressed_content_.data() != header.directory_size)
```

runs only *after* the walk has already read out of bounds, and `directory_size` is itself an unvalidated header field, so a malformed file can simply be built to satisfy it.

This is reachable today from any corrupt or truncated `.aks2` — a partial download or a bad disk is enough to crash the process rather than fall back cleanly.

## Changes

All in `v3src/packed_kernel_common.cc`, plus two default member initialisers in `include/aotriton/_internal/packed_kernel.h`.

- **Validate the header before allocating.** Reject `uncompressed_size` of zero or above 1 GiB, `directory_size` larger than `uncompressed_size`, and `number_of_kernels` larger than the directory can physically hold. The bounds come from the format itself (see `python/aks2.py`): the payload is the directory followed by the images, and each entry costs at least `sizeof(AKS2_Metadata)` plus its name's trailing `\0`.
- **Catch `bad_alloc` around the `resize`.** It otherwise escapes the constructor through `make_shared()` in `open()` and terminates the process.
- **Require the decompressed stream to be exactly `uncompressed_size` bytes.** A short stream previously left the tail of the buffer zero-filled and parsed it as directory content.
- **Bound every directory entry** against the directory region; require `filename_length` to be non-zero and actually NUL-terminated, and build the `string_view` with an explicit length rather than relying on `strlen` finding a terminator.
- **Validate `offset`/`image_size` against the image region at parse time**, so `filter()` cannot hand an out-of-range pointer to `hipModuleLoadDataEx()`.
- **Give the LZMA decoder a finite memlimit** instead of `UINT64_MAX`.
- **Call `lzma_end()`.** It is currently never called — every AKS2 load leaks decoder state, including successful ones. Handled with a small RAII guard because the constructor has several exit paths.

Valid AKS2 files parse identically. The only behaviour change is that malformed ones are rejected with a status instead of read out of bounds.

## Validation

An ASan+UBSan harness drives the real constructor over a corpus of malformed AKS2 files (the harness links `packed_kernel_common.cc`, `fd_unix.cc` and `log.cc`, stubs `lszip`, and reads every byte `filter()` reports as a loadable image, so an out-of-range `offset`/`image_size` surfaces as an ASan report rather than a silently wild pointer).

| case | before | after |
|---|---|---|
| `valid` | accepted + **leak** | accepted, no findings |
| `bad_magic` | rejected | rejected |
| `nkernels_overflow` | **heap-buffer-overflow** | rejected |
| `unterminated_name` | **heap-buffer-overflow** | rejected |
| `image_size_oob` | **heap-buffer-overflow** | rejected |
| `offset_oob` | **SEGV** | rejected |
| `zero_uncompressed` | **SEGV** | rejected |
| `filename_len_oob` | leak | rejected |
| `dirsize_gt_total` | leak | rejected |
| `truncated_stream` | leak | rejected |
| `huge_uncompressed` | leak (4 GiB alloc) | rejected |

The valid file decodes byte-identically before and after (same image sizes, same checksums).

600 random header/body mutations of a valid file produce no sanitizer findings against this branch; the same mutation run against current `main` hits a heap-buffer-overflow within 11 iterations.

I kept the harness out of the diff since the repo has no C++ test infrastructure and wiring some up is a separate decision — happy to add it (or a `LLVMFuzzerTestOneInput` entry point) if you'd like it in tree.

## Notes for reviewers

- The 1 GiB and 256 MiB constants are deliberately loose — they exist to bound what a corrupt `u32` can request, not to reflect real entry sizes. Say the word if you'd prefer different values or a build-time knob.
- `bad_magic` already fails on `main`; it's in the corpus as a regression guard for the check fixed in d1e09cc (#177).
- Reported internally as SWSPLAT-23886, which was filed against a pre-#177 vendored snapshot and described the magic-check bug as remotely exploitable. That part does not apply to `main` — the image directory is `dirname(libaotriton.so)/aotriton.images`, so writing there already implies the ability to replace the library itself. The bounds issues above are real regardless, which is what this PR addresses.

### DIFF
--- a/v3src/packed_kernel_common.cc
+++ b/v3src/packed_kernel_common.cc
@@ -10,6 +10,7 @@
 #include <mutex>
 #include <cstring>
 #include <cassert>
+#include <new>
 #if defined(_WIN32)
 #include <windows.h>
 #else
@@ -25,6 +26,14 @@ namespace fs = std::filesystem;
 static const std::string_view KERNEL_STORAGE_V2_BASE = "aotriton.images";
 static const std::string AKS2_MAGIC = "AKS2";
 constexpr int AOTRITON_LZMA_BUFSIZ = 64 * 1024;
+// Upper bound on the decompressed size of a single AKS2 entry. Shipped entries
+// are a few MiB; this only stops a corrupt header from requesting a multi-GiB
+// allocation before any of its content has been validated.
+constexpr uint64_t AOTRITON_AKS2_MAX_UNCOMPRESSED = 1ull << 30; // 1 GiB
+// Memory budget for the LZMA decoder. The largest dictionary xz selects at -9
+// is 64 MiB, so this leaves generous headroom while still bounding the
+// allocation a crafted stream header can request.
+constexpr uint64_t AOTRITON_LZMA_MEMLIMIT = 256ull << 20; // 256 MiB
 
 namespace {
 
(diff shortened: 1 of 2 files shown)


v3src/packed_kernel_common.cc
```diff
@@ -263,23 +301,60 @@ PackedKernel::PackedKernel(fd_t fd, size_t offset, size_t size) {
   }
   AOTRITON_LOG(LOG_DEBUG, "PackedKernel decompressed to %p",
                static_cast<const void*>(decompressed_content_.data()));
-  const uint8_t* parse_ptr = decompressed_content_.data();
+  auto reject = [this](hipError_t status) {
```
xinyazhang, line 304: Let's change the pattern a little bit: instead of calling reject at every early exit. Let's use temporary  `decompressed_content` and `directory` throughout the processing, and only move them to the member variable `decompressed_content_` `directory_` when everything is okay.

v3src/packed_kernel_common.cc
```diff
@@ -263,23 +301,60 @@ PackedKernel::PackedKernel(fd_t fd, size_t offset, size_t size) {
...
+    directory_.clear();
+    final_status_ = status;
+  };
+  // A short stream would leave the tail of the buffer zero-filled and parsed as
+  // if it were real directory content.
+  if (strm.total_out != header.uncompressed_size) {
+    AOTRITON_LOG(LOG_DEBUG, "AKS2 payload is %llu bytes, header declares %u",
+                 static_cast<unsigned long long>(strm.total_out),
+                 unsigned(header.uncompressed_size));
+    reject(hipErrorIllegalState);
```
xinyazhang, line 315: With the new pattern about this can be `final_status_ = hipErrorIllegalState; return`

xinyazhang (changes requested):
Overall LGTM, but the `reject` part can be adjusted a little.

PR #216, opened 2026-08-18
Restore flash kernel tuning config generators in ATI

nemanjaudovic:
# Overview

Restore the flash kernel tuning configuration generators left as single-config
placeholders during the ATI migration.

## Changes

* [flash,tune] Restore the release/0.12 candidate spaces and ordering for
`attn_fwd`, `bwd_kernel_dk_dv`, `bwd_kernel_dq`, and `bwd_kernel_fuse`.
* [test] Add CPU-only ordered parity and Config-to-KernelSignature tests using
the real ATI descriptions.

## Testing

* Focused autotune-config suite: 23 passed.
* Full CPU-only self-test: 237 passed, 26 skipped.

## AI Disclosure

* Code was generated with help of Claude Code.
* All of the code was reviewed by human.

### DIFF
--- a/modules/flash/aot/attn_fwd.py
+++ b/modules/flash/aot/attn_fwd.py
@@ -11,12 +11,15 @@ instantiation description onto it. Covers all 74 parameters. Conditional argumen
 the derive fixes the per-functional value (ati+newbinds_rev1.md §6.2).
 """
 
+import itertools
+import os
 from dataclasses import dataclass
 
 import numpy as np
 
 import aotriton.template_instantiation as ati
-from ._common import flash_disabled, block_dmodel_values, MAIN_DTYPES
+from aotriton.gpu_targets import AOTRITON_ARCH_WARPSIZE
+from ._common import flash_disabled, block_dmodel_values, MAIN_DTYPES, check_value
 
 
 @dataclass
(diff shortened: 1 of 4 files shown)


xinyazhang (changes requested):
`python/test/test_flash_real_gen_autotune_configs.py` does not belong to `python/test/`. Should be located under `modules/flash`.

However a more critical question is "what does it test against"? The `gen_autotune_configs` are supplying data only. Writing UTs for those functions is like to write UTs for a dict

nemanjaudovic:
Removed the test. The PR is now just the four generator files.

Agreed on the substance: the config list is the spec, so asserting it matches a second copy of itself proves little.  Adding unit tests for this purpose seems not portable across version bumps/config changes. I verified the port out-of-tree instead.

1. Differential against release/0.12. The script below (I got it with help of Claude) AST-extracts gen_autotune_configs from git show release/0.12:v3python/rules/flash/*.py and executes it against a shim, comparing config-for-config with the current generators driven by real Functional objects from the linker. The reference side is the actual historical code, not a copy of it.

The generators read only arch, dtype, BLOCK_DMODEL and CAUSAL_TYPE, so this covers their whole input domain:

    attn_fwd              1080 combinations  OK
    bwd_kernel_dk_dv      1080 combinations  OK
    bwd_kernel_dq          1080 combinations  OK
    bwd_kernel_fuse         990 combinations  OK

    compared : 4230 combinations
    mismatch : 0

bwd_kernel_fuse is compared against its perf schema only: 0.12 pushed NUM_XCDS through update_programmatic_perfs for it and the signature builder dropped it again; ATI just never emits it.

2. AOTRITON_BUILD_FOR_TUNING=ON codegen, five architectures (NOIMAGE_MODE), which runs gen_signatures_for_tuning() over every config of every functional. All generated clean:

                 attn_fwd   dk_dv      dq    fuse
    gfx90a          29760   16128   17280   14784
    gfx942          29760   16128   17280   14784
    gfx950          30432   13440   14400   12096
    gfx1100         21120   15840   15840   14256
    gfx1201         23040   17280   17280   14256

Those counts are reproducible without the build: a second script walks the linked kernel descriptions, skips disabled functionals with the same is_functional_disabled predicate codegen/kernel.py:26 uses, sums len(gen_autotune_configs(f)) per kernel, and diffs against the build's Bare.compile.
Exact match, so the emitted compile list is precisely the generator output - nothing dropped or duplicated in between (gfx1201 shown):

    kernel                            computed   emitted
    attn_fwd                             23040     23040      OK
    bwd_kernel_dk_dv                     17280     17280      OK
    bwd_kernel_dq                        17280     17280      OK
    bwd_kernel_fuse                      14256     14256      OK
    bwd_preprocess                          72        72      OK
    bwd_preprocess_varlen                   72        72      OK
    debug_simulate_encoded_softmax           3         3      OK
    ----------------------------------------------------------
    TOTAL                                72003     72003      OK

  The last three are untunable - one default signature per functional - and unaffected by this PR.

  Build log:
  
[build.txt](https://github.com/user-attachments/files/31225007/build.txt)

Script used for comparing generator with 0.12:
[verify_generator_parity.py](https://github.com/user-attachments/files/31225030/verify_generator_parity.py)

Script used to reproduce the emitted config counts directly from the generator[verify_config_counts.py](https://github.com/user-attachments/files/31225032/verify_config_counts.py)

xinyazhang (approved):
LGTM

PR #221, opened 2026-08-26
Handle failed kernels in tuned builds

nemanjaudovic:
## Summary

Prevent database-backed dispatch tables from selecting Triton kernels whose
build produced an unusable artifact. With a positive
`AOTRITON_GPU_BUILD_TIMEOUT`, `compile.py` may record a failure as a zero-byte
`.hsaco` plus status JSON and still let the build continue; that image was then
packaged and could fail at runtime with `hipErrorInvalidImage`.

## Implementation

After HSACO compilation, rerun code generation and treat a signature as usable
only when its image is nonempty and its status is `Complete`.

For a database-backed functional, repoint failed LUT entries to the compiled
signature already serving the most cells. This is a generic, deterministic
fallback that needs no kernel-specific performance-model knowledge. Keep the
signature list unchanged because its order defines offsets into the packed
string table in the first-pass `shim.<kernel>.cc`. Also retain the first-pass
LUT integer type because that shim is not regenerated.

Fail with `NoCompiledKernel` when no compiled signature survives. A functional
built with only its default kernel has no replacement, so it emits a warning and
keeps the existing behavior.

The second pass leaves `Bare.*` files and tuning manifests untouched. CMake
models its tuning-table `.cc` files as generated outputs depending on all HSACOs,
so Ninja recompiles changed tables in the same invocation. A stamp reruns the
second pass after reconfiguration, when the configure-time first pass may have
restored the original tables. No-image builds skip the pass.

This handles recorded compile failures only; it cannot detect a nonempty kernel
that compiles successfully but is incorrect at load or execution time.

## Testing

- CPU-only focused tests: artifact validation, LUT repointing, signature/order and
  LUT-type preservation, no-survivor and default-only behavior, build-mode
  routing, and preservation of generated manifests.
- Full Python suite: `279 passed, 2 skipped`.
- Manual 19,187-kernel gfx1100 build: healthy pass was a no-op; one failed kernel
  changed and rebuilt one tuning table; reconfigure reran the repair; the next
  Ninja invocation reported no work.

## AI Disclosure
  Help of Claude Code was used for writing this code.
  All the code was reviewed and tested by human.

### DIFF
--- a/python/codegen/autotune.py
+++ b/python/codegen/autotune.py
@@ -15,11 +15,13 @@ from ..utils import (
 from .common import (
     codegen_struct_cfields,
     MissingLutEntry,
+    NoCompiledKernel,
     hsaco_ondisk_name,
     hsaco_dir,
 )
 from .basetune import BaseTuneCodeGenerator
 import json
+import sys
 import numpy as np
 
 class AutotuneCodeGenerator(BaseTuneCodeGenerator):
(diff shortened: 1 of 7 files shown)


xinyazhang (changes requested):
The signature is already excluded with current code
https://github.com/ROCm/aotriton/blob/30ae96716b3e7a1bf0d7892ce3b8288244f8ebec/python/codegen/autotune.py#L59

And those entries will not be included in tuning database. Any build without current tunning database + pinned Compiler is unsupported.

xinyazhang:
Do not 100% trust the plan or bug given by the AI agent. Length bias is [a well-known problem for LLM](https://arxiv.org/abs/2310.03716) and still applies to frontier models.

Specific to this PR. A tuning database entry is broken if it cannot guide the compiler to generate .hsaco file.
For tuning mode this is addressed by the `hsaco_compile_successful` function.
For testing the compiler will raise error to the build system because they take `--error_when_timeout`.

Close unless you can reproduce the anticipated bug.

nemanjaudovic:
> The signature is already excluded with current code
> 
> https://github.com/ROCm/aotriton/blob/30ae96716b3e7a1bf0d7892ce3b8288244f8ebec/python/codegen/autotune.py#L59
> 
> And those entries will not be included in tuning database. Any build without current tunning database + pinned Compiler is unsupported.

This PR mainly targeted problems that could arise from configs reusage (specifically gfx11x are reusing gfx1100 configs) - not tuning builds problems. 
If config search space for gfx1100 is widened (as in https://github.com/ROCm/aotriton/pull/220), some more 'liberal' configs can win and compile successfully for gfx1100 (like with bigger tiles, or whatever) - and end up in db. Since other gfx11x are reusing that db, I created this PR to address problems like this - config compiles successfully for gfx1100 but fails on gfx1101/1102... 
With current code (without https://github.com/ROCm/aotriton/pull/220), that problem is less likely to happen, since configs that are being searched are in 'safe zone' even for other gfx.
I planned to leave this PR in Draft, to see how results of (hopefully) widened config tuning are, and then if this problem becomes real to investigate it further. 
Also, i am still not sure about is this the best mechanism for this kind of problem, but will investigate further. Wanted to create this Draft PR to see if this makes sense at all, since I wanted to propose wider config sets for gfx1100, i didnt want other problems to arise from that...

PR #228, opened 2026-09-14
fixed bias and alibi_block f32 asymmtry between the fwd and bwd

Chi-Chu319:
There was asymmetry between the fwd and bwd calculation of bias * log_2^e. The forward has it in bf16 and the backward has it in f32. Now the asymmetry is removed


xinyazhang (approved):
LGTM.

